PR #32, opened 2023-03-23
Metrics: F1Beta, MCC

jegorus:
RT-6. Added f1beta, mcc and tests

### DIFF
--- a/rectools/metrics/classification.py
+++ b/rectools/metrics/classification.py
@@ -18,6 +18,7 @@ import typing as tp
 from collections import defaultdict
 
 import attr
+import numpy as np
 import pandas as pd
 
 from rectools import Columns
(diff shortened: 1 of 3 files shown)


rectools/metrics/classification.py
```diff
@@ -424,3 +425,75 @@ def make_confusions(reco: pd.DataFrame, interactions: pd.DataFrame, k: int) -> p
     merged = merge_reco(reco, interactions)
     confusion_df = calc_confusions(merged, k)
     return confusion_df
+
+
+@attr.s
+class F1Beta(SimpleClassificationMetric):
+    """
+    Calculates metric base on precision@k and recall@k
```
feldlime, line 433:
```suggestion
    Fbeta score for k first recommendations.
```

rectools/metrics/classification.py
```diff
@@ -424,3 +425,75 @@ def make_confusions(reco: pd.DataFrame, interactions: pd.DataFrame, k: int) -> p
...
+        - ``fp`` - number of non-relevant recommendations among the first `k` items of recommendation list;
+        - ``fn`` - number of items the user has interacted with but that weren't recommended (in top-`k`).
+
+    Parameters
+    ----------
+    k : int
+        Number of items in top of recommendations list that will be used to calculate metric.
+    """
+
+    def _calc_per_user_from_confusion_df(self, confusion_df: pd.DataFrame, catalog: Catalog) -> pd.Series:
```
feldlime, line 490: why do you need `catalog` here? And why don't you use `SimpleClassificationMetric` as the base class?

feldlime (changes requested):

rectools/metrics/classification.py
jegorus: I need it to use `TN`, which is calculated in `ClassificationMetric` using `catalog`:
`confusion_df[TN] = len(catalog) - self.k - confusion_df[FN]` (from function `calc_per_user_from_confusion_df`)

rectools/metrics/classification.py
jegorus: changed

rectools/metrics/classification.py
feldlime: ok

sharthZ23:
Approve

PR #33, opened 2023-03-23
model_selection/last_n_split

yukeeul:
Добавил возможность отделить последние n взаимодействий каждого пользователя для теста + тесты


rectools/model_selection/last_n_split.py
```diff
@@ -0,0 +1,121 @@
...
+        for n in self.n:
+            if n <= 0:
+                raise ValueError(f"N must be positive, got {n}")
+
+            grouped_df = df.groupby("user_id")["datetime"].nlargest(n)
+            test_interactions = grouped_df.keys().to_numpy()
+            get_second_value = np.vectorize(lambda x: x[1])
+            test_interactions = get_second_value(test_interactions)
+            test_mask = np.zeros_like(idx, dtype=bool)
+            test_mask[test_interactions] = True
```
feldlime, line 111: 1. Too difficult and slow, use `i.levels[1].values` to get values of 1st level of `MultiIndex`, also use `.index` instead of `.keys()`
2. Bad naming: `grouped_df` isn't a DataFrame, `test_interactions` isn't an interactions. Variable names have to be informative
3. Will work incorrectly if index of `interactions.df` is not [0, 1, 2, ...] (spoil default index in tests please)

feldlime (changes requested):

rectools/model_selection/last_n_split.py
```diff
@@ -0,0 +1,121 @@
...
+class LastNSplitter(Splitter):
+    """
+    Generate train and test putting last n interaction for
+    each user in test and others in train.
+    It is also possible to exclude cold users and items
+    and already seen items.
+
+    Parameters
+    ----------
+    n: int or iterable of ints
```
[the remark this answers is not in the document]
yukeeul: Done

rectools/model_selection/last_n_split.py
```diff
@@ -0,0 +1,121 @@
...
+        filter_cold_users: bool = True,
+        filter_cold_items: bool = True,
+        filter_already_seen: bool = True,
+    ) -> None:
+        super().__init__(filter_cold_users, filter_cold_items, filter_already_seen)
+        self.n: tp.Iterable[int]
+        if isinstance(n, int):
+            self.n = [n]
+        else:
+            self.n = n
```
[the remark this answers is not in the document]
yukeeul: Done

rectools/model_selection/last_n_split.py
```diff
@@ -0,0 +1,121 @@
...
+    """
+    Generate train and test putting last n interaction for
+    each user in test and others in train.
+    It is also possible to exclude cold users and items
+    and already seen items.
+
+    Parameters
+    ----------
+    n: int or iterable of ints
+        Number of interactions for each user that will be included in test
```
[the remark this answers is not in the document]
yukeeul: Done

rectools/model_selection/last_n_split.py
yukeeul: Done

rectools/model_selection/last_n_split.py
```diff
@@ -0,0 +1,118 @@
...
+
+        for n in self.n:
+            if n <= 0:
+                raise ValueError(f"N must be positive, got {n}")
+
+            last_n_interactions = df.groupby("user_id")["datetime"].nlargest(n)
+            test_idx = last_n_interactions.index.levels[1].to_numpy()
+            test_mask = np.zeros_like(idx, dtype=bool)
+            test_mask[test_idx] = True
+            train_mask = ~test_mask
```
feldlime, line 111: Just a bit shorter

```suggestion
            train_mask = np.zeros_like(idx, dtype=bool)
            train_mask[test_idx] = False
```

yukeeul: Done

rectools/model_selection/kfold_split.py
```diff
@@ -124,12 +124,9 @@ def _split_without_filter(
...
             fold_info = {"fold_number": i}
-            test_mask = np.zeros_like(idx, dtype=bool)
-            chosen_idx = shuffled_idx[i * test_part_size : (i + 1) * test_part_size]
-            test_mask[chosen_idx] = True
-            train_mask = ~test_mask
-
+            train_mask = np.ones_like(idx, dtype=bool)
+            test_idx = shuffled_idx[i * test_part_size : (i + 1) * test_part_size]
+            train_mask[test_idx] = False
             train_idx = idx[train_mask].values
```
feldlime, line 130: We can take it a step further

```suggestion
            left = i * test_part_size
            right = (i + 1) * test_part_size
            test_idx = shuffled_idx[left : right]
            train_idx = np.concatenate((shuffled_idx[:left], shuffled_idx[right:]))
```

It's a bit simpler and approx twice faster. But `train_idx` will be shuffled

rectools/model_selection/last_n_split.py
feldlime: Tried couple of tests and found out that current code does not work when `df.index` is not a permutation of [0, 1, 2...len(df)-1] (e.g.  if `df.index` contains values bigger that len(df) or some values are missed, or there are duplicates).

So let's please rewrite it and add tests for all such cases.

And I'm pretty sure that using the original index in any way is a bad idea. Or we need to add constraints for it in `Interactions`. But now there are none of them

rectools/model_selection/kfold_split.py
yukeeul: Done

rectools/model_selection/last_n_split.py
yukeeul: I've added tests for such cases.  However, I didn't find a solution that is fast and works correctly with duplicates in `df.index` so as a temporary decision slow code is used only in this specific case

feldlime (approved):

PR #35, opened 2023-03-30
Metrics. Roc-curve plot and auc added

jegorus:
Implemented roc-curve and auc, but have some questions:
1) It seems that roc for k extends plot linearly and connect with (1, 1) like: [https://wiki.epfl.ch/edicpublic/documents/Candidacy%20exam/Evaluation.pdf]
But usually it is implemented using scores for all items, but it seems that modelBase doesn't do this. If I tried to use k = len(catalog) (for all elements) some of them were discarded (about 30-50% for user) I've tried debugging, but I'm not sure if it's supposed to work like this.
2) For AUC I've put scores for not recommended elements as 0 (same reason (1)). So it works like LAUC from article from (1)
3) I optimized roc-plot by changing confusion matrix instead of recalculating using make_confusions and went from 15 sec to 1 sec on ratings.dat, but AUC works really slow: more than 30 sec for user. I've tried to use numpy vectorize and broadcasting, but it didn't work because of addressing pandas. I'll try to find a way to do this, but maybe you have suggestions
4) I'm not sure how to test the plot, but tested tpr/fpr calculator. Also I added test for one user auc but didn't add test for all users because of speed.
5) linter passed flake8, but gives codespelling mistake, because it doesn't recognize fpr. I'm not sure if I had to change it in smth like makefile. Maybe you faced similar problem

ROC-curve screenshot:
ROC-curve screenshot:

### DIFF
--- a/rectools/metrics/__init__.py
+++ b/rectools/metrics/__init__.py
@@ -52,6 +52,7 @@ from .novelty import MeanInvUserFreq
 from .ranking import MAP, MRR, NDCG
 from .scoring import calc_metrics
 from .serendipity import Serendipity
+from .roc_auc import AUC
 
 __all__ = (
     "Precision",
@@ -62,6 +63,7 @@ __all__ = (
     "MAP",
     "NDCG",
     "MRR",
+    "AUC",
     "MeanInvUserFreq",
     "IntraListDiversity",
     "Serendipity",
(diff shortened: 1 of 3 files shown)


jegorus:
6) Tests are failed here because it cannot find matplotlib. I'm not sure if I should add it to requirements
7) linter found isort problem I fixed it. I'll commit it with next changes

PR #38, opened 2023-05-04
Coverage metrics

jegorus:
Item coverage and num retrieved

### DIFF
--- a/rectools/metrics/coverage.py
+++ b/rectools/metrics/coverage.py
@@ -1,0 +1,109 @@
+import attr
+import pandas as pd
+
+from rectools import Columns
+
+from .base import Catalog, MetricAtK
+
+
+@attr.s
+class ItemCoverage(MetricAtK):
+    """
+    Item space coverage (also referred as catalog coverage) is a metric that shows
+    what part of the items is covered by first k recommendations
+    ItemCoverage = #recommended_items / num_items
+
+    Parameters
+    ----------
+    k : int
+        Number of items in top of recommendations list that will be used to calculate metric.
+
+    """
+
+    def calc(self, reco: pd.DataFrame, catalog: Catalog) -> float:
+        """
+        Calculate item space coverage for all users
+
+        Parameters
+        ----------
+        reco : pd.DataFrame
+            Recommendations table with columns `Columns.User`, `Columns.Item`, `Columns.Rank`.
+        catalog : collection
+            Collection of unique item ids that could be used for recommendations.
+
+        Returns
+        -------
+        float
+            Value of metric.
+        """
+        reco_k_first_ranks = reco[reco[Columns.Rank] <= self.k]
+        return len(reco_k_first_ranks[Columns.Item].unique()) / len(catalog)
+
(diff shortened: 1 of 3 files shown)


rectools/metrics/coverage.py
```diff
@@ -0,0 +1,109 @@
...
+            Recommendations table with columns `Columns.User`, `Columns.Item`, `Columns.Rank`.
+        catalog : collection
+            Collection of unique item ids that could be used for recommendations.
+
+        Returns
+        -------
+        float
+            Value of metric.
+        """
+        reco_k_first_ranks = reco[reco[Columns.Rank] <= self.k]
```
feldlime, line 39: As we need only the items column, let's take only it.

This will be more memory efficient and faster as well
```
items = reco.loc[reco[Columns.Rank] <= self.k, Columns.Item]
```

rectools/metrics/coverage.py
```diff
@@ -0,0 +1,109 @@
...
+        return len(reco_k_first_ranks[Columns.Item].unique()) / len(catalog)
```
feldlime, line 40: There is `nunique` method, no need to use `len`

rectools/metrics/coverage.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def calc_per_user(self, reco: pd.DataFrame, catalog: Catalog) -> pd.Series:
```
feldlime, line 42: Maybe for this metric `calc_per_user` method is meaningless since recommendations for user are unique and coverage depends only on `k` that we set up here

rectools/metrics/coverage.py
```diff
@@ -0,0 +1,109 @@
...
+        ----------
+        reco : pd.DataFrame
+            Recommendations table with columns `Columns.User`, `Columns.Item`, `Columns.Rank`.
+
+        Returns
+        -------
+        pd.Series
+            Values of metric (index - user id, values - metric value for every user).
+        """
+        reco_k_first_ranks = reco[reco[Columns.Rank] <= self.k]
```
feldlime, line 108: Same here

rectools/metrics/coverage.py
```diff
@@ -0,0 +1,109 @@
...
+        return reco_k_first_ranks.groupby(Columns.User)[Columns.Item].count().rename(None)
```
feldlime, line 109: It's better to store results of complex expressions to the separate variables

feldlime, line 109: It's better to store results of complex expressions to the separate variables

feldlime (changes requested):

rectools/metrics/coverage.py
```diff
@@ -0,0 +1,109 @@
...
+        """
+        reco_k_first_ranks = reco[reco[Columns.Rank] <= self.k]
+        return reco_k_first_ranks.groupby(Columns.User)[Columns.Item].nunique().rename(None) / len(catalog)
+
+
+@attr.s
+class NumRetrieved(MetricAtK):
+    """
+    Number of recommendations retrieved is a metric that shows
+    how much items were recommended to users by first k recommendations (less or equal k)
```
feldlime, line 66: `much` -> `many`

PR #49, opened 2023-10-10
Support for fit_features_together for implicit als

blondered:
- Added support for `fit_features_together`=True for implicit ALS model wrapper

### DIFF
--- a/rectools/models/implicit_als.py
+++ b/rectools/models/implicit_als.py
@@ -19,14 +19,15 @@ import implicit.gpu
 import numpy as np
 from implicit.cpu.als import AlternatingLeastSquares as CPUAlternatingLeastSquares
 from implicit.gpu.als import AlternatingLeastSquares as GPUAlternatingLeastSquares
+from implicit.utils import check_random_state
 from scipy import sparse
+from tqdm.auto import tqdm
 
 from rectools.dataset import Dataset, Features
 from rectools.exceptions import NotFittedError
 
 from .vector import Distance, Factors, VectorModel
 
-MAX_GPU_FACTORS = 1024
 AVAILABLE_RECOMMEND_METHODS = ("loop",)
 AnyAlternatingLeastSquares = tp.Union[CPUAlternatingLeastSquares, GPUAlternatingLeastSquares]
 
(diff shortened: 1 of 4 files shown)


rectools/models/implicit_als.py
```diff
@@ -218,8 +212,237 @@ def _fit_paired_factors(
...
+            }
+        )
         features_model = CPUAlternatingLeastSquares(**features_model_params)
         features_model.item_factors = y_factors
         features_model.fit(xy_csr)
         x_factors = features_model.user_factors
     return x_factors
+
+
+def _init_latent_factors_cpu(
```
feldlime, line 229: Please add to this and to the following function short comments that the logic is copied and pasted from the implicit

tests/models/test_implicit_als.py
```diff
@@ -105,6 +105,38 @@ def test_basic(
             actual,
         )
 
+    @pytest.mark.parametrize("fit_features_together", (False, True))
+    def test_consistent_with_pure_implicit(self, dataset: Dataset, fit_features_together: bool, use_gpu: bool) -> None:
+        base_model = AlternatingLeastSquares(factors=10, num_threads=2, iterations=30, use_gpu=use_gpu)
+        users = np.array([10, 20, 30, 40])
+        self._init_model_factors_inplace(base_model, dataset)
```
feldlime, line 112: looks like a hack

Why not just fix `random_state`?

tests/models/test_implicit_als.py
```diff
@@ -105,6 +105,38 @@ def test_basic(
...
+        actual_reco = wrapped_model.recommend(
+            users=users,
+            dataset=dataset,
+            k=3,
+            filter_viewed=False,
+        )
+
+        ui_csr = dataset.get_user_item_matrix(include_weights=True)
+        base_model.fit(ui_csr)
+        for user_id in users:
```
feldlime, line 126: Isn't it correct that implicit can recommend for multiple users? Why do you need the cycle?

blondered: I believe it's just simpler to match recos from different methods this way

rectools/models/implicit_als.py
blondered: Done

feldlime (approved):

PR #69, opened 2023-11-26
Feature/interactions to raw

feldlime:
- Added methods for conversion `Interactions` to raw form and for getting raw interactions from `Dataset`

### DIFF
--- a/rectools/dataset/interactions.py
+++ b/rectools/dataset/interactions.py
@@ -125,8 +125,6 @@ class Interactions:
         """
         Form an user-item CSR matrix based on `interactions` attribute.
 
-        It is used `Interactions.get_user_item_matrix`, see its documentations for details.
-
         Parameters
         ----------
         include_weights : bool, default ``True``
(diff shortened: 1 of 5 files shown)


rectools/dataset/interactions.py
```diff
@@ -125,8 +125,6 @@ def get_user_item_matrix(self, include_weights: bool = True) -> sparse.csr_matri
         """
         Form an user-item CSR matrix based on `interactions` attribute.
```
blondered, line 126:
```suggestion
        Form a user-item CSR matrix based on interactions data.
```

rectools/dataset/interactions.py
blondered, line 1: Let's also fix docstring for `Interactions` class. https://github.com/MobileTeleSystems/RecTools/blob/a7e6407a7a969f72d22f2806eea74687b9b82ece/rectools/dataset/interactions.py#L30  `Structure to store info about user-item interactions.`

rectools/dataset/interactions.py
```diff
@@ -152,3 +150,42 @@ def get_user_item_matrix(self, include_weights: bool = True) -> sparse.csr_matri
...
+        Convert itself to `pd.DataFrame` with replacing internal user and item ids to external ones.
+
+        Parameters
+        ----------
+        user_id_map : IdMap
+            User id map that has to be used for converting internal user ids to external ones.
+        item_id_map : IdMap
+            Item id map that has to be used for converting internal item ids to external ones.
+        include_weight : bool, default ``True``
+            Whether include weight column into resulting dataset or not.
```
blondered, line 171: Needed both here, for next attribute and for dataset class.
```suggestion
            Whether to include weight column into resulting table or not.
```

rectools/dataset/interactions.py
blondered, line 1: Can we please fix the https://github.com/MobileTeleSystems/RecTools/blob/a7e6407a7a969f72d22f2806eea74687b9b82ece/rectools/dataset/interactions.py#L73

blondered (changes requested):

blondered:
Let's rename `to_raw` -> `to_external`

rectools/dataset/interactions.py
feldlime: Fixed

rectools/dataset/interactions.py
feldlime: Fixed

rectools/dataset/interactions.py
feldlime: Fixed

rectools/dataset/interactions.py
feldlime: Let's do it in a separate PR

blondered (approved):

PR #71, opened 2023-11-28
Feature/cross validate

feldlime:
- Added `rectools.model_selection.cross_validate` function

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Ability to pass internal ids to `recommend` and `recommend_to_items` methods and get internal ids back ([#70](https://github.com/MobileTeleSystems/RecTools/pull/70))
+- `rectools.model_selection.cross_validate` function ([#71](https://github.com/MobileTeleSystems/RecTools/pull/71))
 
 - Ability to pass internal ids to `recommend` and `recommend_to_items` methods and get internal ids back ([#70](https://github.com/MobileTeleSystems/RecTools/pull/70))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## Unreleased
+
+### Added
+
+- Ability to pass internal ids to `recommend` and `recommend_to_items` methods and get internal ids back ([#70](https://github.com/MobileTeleSystems/RecTools/pull/70))
+
+
 ## [0.4.1] - 31.10.2023
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Ability to pass internal ids to `recommend` and `recommend_to_items` methods and get internal ids back ([#70](https://github.com/MobileTeleSystems/RecTools/pull/70))
 - `rectools.model_selection.cross_validate` function ([#71](https://github.com/MobileTeleSystems/RecTools/pull/71))
 
-- Ability to pass internal ids to `recommend` and `recommend_to_items` methods and get internal ids back ([#70](https://github.com/MobileTeleSystems/RecTools/pull/70))
+### Fixed
+- Small bug in `LastNSplitter` with incorrect `i_split` in info ([#70](https://github.com/MobileTeleSystems/RecTools/pull/70))
 
 
 ## [0.4.1] - 31.10.2023
--- a/rectools/model_selection/cross_validate.py
+++ b/rectools/model_selection/cross_validate.py
@@ -1,0 +1,77 @@
+import typing as tp
+
+from rectools.columns import Columns
+from rectools.dataset.dataset import Dataset
+from rectools.dataset.interactions import Interactions
+from rectools.metrics.base import MetricAtK
+from rectools.metrics.scoring import calc_metrics
+from rectools.model_selection.splitter import Splitter
+from rectools.models.base import ModelBase
+from rectools.types import ExternalIds
+
+
+def cross_validate(  # pylint: disable=too-many-locals
+    dataset: Dataset,
+    splitter: Splitter,
+    metrics: tp.Dict[str, MetricAtK],
+    models: tp.Dict[str, ModelBase],
+    k: int,
+    filter_viewed: bool,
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Ability to pass internal ids to `recommend` and `recommend_to_items` methods and get internal ids back ([#70](https://github.com/MobileTeleSystems/RecTools/pull/70))
+- `rectools.model_selection.cross_validate` function ([#71](https://github.com/MobileTeleSystems/RecTools/pull/71))
```
blondered, line 13: Let's add here `(need examples update)`

rectools/model_selection/cross_validate.py
```diff
@@ -0,0 +1,77 @@
...
+from rectools.types import ExternalIds
+
+
+def cross_validate(  # pylint: disable=too-many-locals
+    dataset: Dataset,
+    splitter: Splitter,
+    metrics: tp.Dict[str, MetricAtK],
+    models: tp.Dict[str, ModelBase],
+    k: int,
+    filter_viewed: bool,
```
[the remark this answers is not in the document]
feldlime: But this option is for model

E.g. it's ok for your service (say music streaming) to recommend already seen items, so you do not filter them in splitter, but it's not obvious when the quality of the model would be better, if you filter them from recommendations or not.

What we probably should discuss:
 How to give an opportunity to check both `filter_viewed = True` and `filter_viewed = False`. But I think it's not urgent now, we can add this later.

CHANGELOG.md
feldlime: forgot about it

added now

rectools/model_selection/cross_validate.py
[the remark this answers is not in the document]
blondered: Good point. Ok, let's keep it this way

rectools/model_selection/cross_validate.py
```diff
@@ -0,0 +1,139 @@
...
+        Whitelist of external item.
+        If given, only these items will be used for recommendations.
+
+    Returns
+    -------
+    dict
+        Dictionary with structure
+        {
+            "splits": [
+                {"i_split": 1, <split_info>},
```
blondered, line 82: let's start from 0 to avoid confusing people

rectools/model_selection/cross_validate.py
```diff
@@ -0,0 +1,139 @@
...
+                {"i_split": 2, <split_info>},
+                ...
+            ],
+            "metrics": [
+                {"model": "model_1", "i_split": 1, <metrics>},
```
blondered, line 87: let's start from 0 here too

rectools/model_selection/cross_validate.py
```diff
@@ -0,0 +1,139 @@
...
+    split_iterator = splitter.split(interactions, collect_fold_stats=True)
+
+    split_infos = []
+    metrics_all = []
+
+    for train_ids, test_ids, split_info in split_iterator:
+        split_infos.append(split_info)
+
+        interactions_df_train = interactions.df.iloc[train_ids]  # 1x internal
+        fold_dataset = gen_2x_internal_ids_dataset(interactions_df_train, dataset.user_features, dataset.item_features)
```
blondered, line 105: Let's write a comment here why we need to do this: we need to avoid fitting models on sparse matrices with all zero rows/columns => we need to create a fold dataset which contains only hot users and items for current training.
Should we rename the func to something more meaningful? Like "get_filtered_train_dataset". The main goal is to filter, mapping is just unavoidable for this. And possibly comment that mapping is done to new x2 internal ids

tests/model_selection/test_cross_validate.py
```diff
@@ -0,0 +1,262 @@
...
+                {"model": "als", "i_split": 1, "precision@2": 0.375, "recall@1": 0.25},
+            ],
+        }
+
+        assert actual == expected
+
+    def test_fail_with_cold_users(self) -> None:
+        splitter = LastNSplitter(n=1, n_splits=2, filter_cold_users=False)
+
+        with pytest.raises(KeyError):
```
blondered, line 254: I guess it is supposed to be NotImplementedError. And we need to check splitter settings and raise this in the beginning of cross validation. Otherwise it is a very unexpected behaviour.

rectools/model_selection/cross_validate.py
```diff
@@ -0,0 +1,139 @@
...
+    models : dict(str -> ModelBase)
+        Dict of initialized model objects to fit and measure quality,
+        where key is metric name and value is metric object.
+    k : int
+        Derived number of recommendations for every user.
+        For some models actual number of recommendations may be less than `k`.
+    filter_viewed : bool
+        Whether to filter from recommendations items that user has already interacted with.
+    items_to_recommend : array-like, optional, default None
+        Whitelist of external item.
```
blondered, line 73: Whitelist of external item ids

rectools/model_selection/cross_validate.py
```diff
@@ -0,0 +1,139 @@
...
+from .splitter import Splitter
+
+
+def gen_2x_internal_ids_dataset(
+    interactions_internal_df: pd.DataFrame,
+    user_features: tp.Optional[Features],
+    item_features: tp.Optional[Features],
+) -> Dataset:
+    """
+    Make new dataset based on given interactions and features from base dataset.
```
blondered, line 21: I suggest to rename this func and provide reasons for usage in docstring (see other comment below)

rectools/model_selection/cross_validate.py
feldlime: fixed

tests/model_selection/test_cross_validate.py
feldlime: It's supposed to be... Yeah, maybe. But it's not. This error is coming from models and I won't change it now.

What about splitter settings, so, I don't think we should control it here because it's not about cross validation. And it's possible for user to write their own model that support recommendations for cold users

I'll add warning

rectools/model_selection/cross_validate.py
feldlime: fixed

rectools/model_selection/cross_validate.py
```diff
@@ -79,18 +80,25 @@ def cross_validate(  # pylint: disable=too-many-locals
...
-                {"model": "model_1", "i_split": 2, <metrics>},
-                {"model": "model_2", "i_split": 1, <metrics>},
                 ...
             ]
         }
     """
+    if not splitter.filter_cold_users:  # TODO: remove when cold users support added
+        warnings.warn(
+            "Currently models do not support recommendations for cold users. "
+            "Use `filter_cold_users` option only with custom models. "
```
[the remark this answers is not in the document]
feldlime: right, fixed

blondered (approved):

PR #73, opened 2023-11-30
Feature/update examples

feldlime:
- Updated cross-validation example


examples/2_cross_validation.ipynb
blondered, line 1: - Add pd.DataFrame(cv_results["metrics"]).head()
- Add fold borders explanation (start and end - which is included to test and which is excluded)
- Header "Split interactions for CV" -> "Prepare cross-validation splitter"

PR #77, opened 2023-12-10
Warm and cold users support

feldlime:
- supported warm users and items in `Dataset`
- removed `return_external_ids` parameter in `recommend` and `recommend_to_items` methods
- supported cold and warm targets in base model
- supported new dataset in cross validation

The first part of https://github.com/MobileTeleSystems/RecTools/issues/87


rectools/model_selection/cross_validate.py
```diff
@@ -49,6 +60,7 @@ def cross_validate(  # pylint: disable=too-many-locals
     k: int,
     filter_viewed: bool,
     items_to_recommend: tp.Optional[ExternalIds] = None,
+    keep_features_for_hot_only: bool = False,
```
blondered, line 63: Can we move this to be a `splitter` argument and instance attribute? Not a `cross-validate` parameter? Intuition is that all of the splitting logic must be in one place. Together with all the "filter_cold_" and so on.

blondered: Apart from `filter_cold_` is there gonna be `filter_warm_` logic in splitter?
In this case `splitter` should probably work with Dataset, not just Interactions.
On the other hand, there can be an additional DatasetSplitter, which can handle the Dataset. And `keep_features_for_hot_only` belongs there.

What's the plan?

blondered (changes requested):

rectools/dataset/dataset.py
```diff
@@ -59,6 +59,38 @@ class Dataset:
     user_features: tp.Optional[Features] = attr.ib(default=None)
     item_features: tp.Optional[Features] = attr.ib(default=None)
 
+    @property
+    def n_hot_users(self) -> int:
+        """
+        Return number of hot users in dataset.
+        Users with internal ids from `0` to `n_hot_users - 1` are hot (they present in interactions).
```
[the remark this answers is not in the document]
feldlime: fixed

rectools/dataset/dataset.py
```diff
@@ -59,6 +59,38 @@ class Dataset:
...
+        Users with internal ids from `n_hot_users` to `dataset.user_id_map.size - 1` are warm
+        (they don't present in interactions, but they have features).
+        """
+        return self.interactions.df[Columns.User].max() + 1
+
+    @property
+    def n_hot_items(self) -> int:
+        """
+        Return number of hot items in dataset.
+        Items with internal ids from `0` to `n_hot_items - 1` are hot (they present in interactions).
```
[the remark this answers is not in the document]
feldlime: fixed

rectools/dataset/dataset.py
```diff
@@ -59,6 +59,38 @@ class Dataset:
...
+        Items with internal ids from `n_hot_items` to `dataset.item_id_map.size - 1` are warm
+        (they don't present in interactions, but they have features).
```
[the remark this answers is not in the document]
feldlime: fixed

rectools/model_selection/cross_validate.py
feldlime: Discussed separately. Decided to keep it as is for now

blondered, line 63: It still blows my mind :)
Can we name it more meaningful somehow?

`allow_warm_test_inference: bool = True`
`allow_warm_inference: bool = True`
or even
`prefer_warm_inference_over_cold: bool = True`

rectools/model_selection/last_n_split.py
```diff
@@ -45,8 +45,10 @@ class LastNSplitter(Splitter):
         Number of test folds.
     filter_cold_users : bool, default ``True``
         If `True`, users that not in train will be excluded from test.
```
blondered, line 47:
```suggestion
        If `True`, users that are not present in train will be excluded from test.
```

rectools/model_selection/last_n_split.py
```diff
@@ -45,8 +45,10 @@ class LastNSplitter(Splitter):
...
+        WARNING: both cold and warm users will be excluded from test.
     filter_cold_items : bool, default ``True``
         If `True`, items that not in train will be excluded from test.
```
blondered, line 50:
```suggestion
        If `True`, items that are not present in train will be excluded from test.
```

rectools/model_selection/last_n_split.py
```diff
@@ -45,8 +45,10 @@ class LastNSplitter(Splitter):
...
+        WARNING: both cold and warm items will be excluded from test.
     filter_already_seen : bool, default ``True``
         If ``True``, pairs (user, item) that are in train will be excluded from test.
```
blondered, line 53:
```suggestion
        If ``True``, pairs (user, item) that are present in train will be excluded from test.
```

rectools/model_selection/random_split.py
```diff
@@ -45,8 +45,10 @@ class RandomSplitter(Splitter):
         Controls randomness of each fold. Pass an int to get reproducible result across multiple `split` calls.
     filter_cold_users : bool, default ``True``
         If `True`, users that not in train will be excluded from test.
```
blondered, line 47:
```suggestion
        If `True`, users that are not present in train will be excluded from test.
```

rectools/model_selection/random_split.py
```diff
@@ -45,8 +45,10 @@ class RandomSplitter(Splitter):
...
+        WARNING: both cold and warm users will be excluded from test.
     filter_cold_items : bool, default ``True``
         If `True`, items that not in train will be excluded from test.
```
blondered, line 50:
```suggestion
        If `True`, items that are not present in train will be excluded from test.
```

rectools/model_selection/time_split.py
```diff
@@ -57,8 +57,10 @@ class TimeRangeSplitter(Splitter):
         Number of test folds.
     filter_cold_users : bool, default ``True``
         If `True`, users that not in train will be excluded from test.
```
blondered, line 59:
```suggestion
        If `True`, users that are not present in train will be excluded from test.
```

rectools/models/base.py
```diff
@@ -329,9 +336,177 @@ def _get_sorted_item_ids_to_recommend(
...
+        return df
+
+    def _recommend_cold(
+        self, target_ids: np.ndarray, k: int, sorted_item_ids_to_recommend: tp.Optional[np.ndarray]
+    ) -> tp.Union[SemiInternalRecoTriplet, InternalRecoTriplet]:
+        raise NotImplementedError()
+
+    def _recommend_u2i_warm(
+        self,
+        user_ids: np.ndarray,
```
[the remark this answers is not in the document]
feldlime: I'd like to do this, but we'll encounter the `mypy` error (a good one) because inside this (and others) function we assume it's a numpy array and we use its methods.

If you know a good way to avoid it, let's change.

But anyway, let's do it in a separate PR

rectools/models/base.py
```diff
@@ -329,9 +336,177 @@ def _get_sorted_item_ids_to_recommend(
...
+        )
+
+        if add_rank_col:
+            df[Columns.Rank] = df.groupby(target_col, sort=False).cumcount() + 1
+
+        return df
+
+    def _recommend_cold(
+        self, target_ids: np.ndarray, k: int, sorted_item_ids_to_recommend: tp.Optional[np.ndarray]
+    ) -> tp.Union[SemiInternalRecoTriplet, InternalRecoTriplet]:
```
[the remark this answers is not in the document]
feldlime: you're right, fixed

rectools/models/base.py
```diff
@@ -329,9 +336,177 @@ def _get_sorted_item_ids_to_recommend(
...
+
+class FixedColdRecoModelMixin:
+    """
+    Mixin for models that have fixed cold recommendations.
+
+    Models that use this mixin should implement `_get_cold_reco` method.
+    """
+
+    def _recommend_cold(
+        self, target_ids: np.ndarray, k: int, sorted_item_ids_to_recommend: tp.Optional[np.ndarray]
```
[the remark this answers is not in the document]
feldlime: Similar problem as I described above: it's always `np.ndarray` and we can rely on it but `mypy` will raise an error if you do smth like `target_ids[mask]`

rectools/models/base.py
```diff
@@ -230,63 +249,65 @@ def recommend_to_items(
...
-    def _recommend_u2i(
-        self,
-        user_ids: np.ndarray,
-        dataset: Dataset,
-        k: int,
-        filter_viewed: bool,
-        sorted_item_ids_to_recommend: tp.Optional[np.ndarray],
-    ) -> tp.Tuple[InternalIds, InternalIds, Scores]:
-        raise NotImplementedError()
+        if filter_itself:
```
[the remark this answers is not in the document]
feldlime: It's faster and more memory effective to do this with internal ids

rectools/dataset/dataset.py
```diff
@@ -59,6 +59,38 @@ class Dataset:
...
+        Users with internal ids from `n_hot_users` to `dataset.user_id_map.size - 1` are warm
+        (they aren't present in interactions, but they have features).
+        """
+        return self.interactions.df[Columns.User].max() + 1
+
+    @property
+    def n_hot_items(self) -> int:
+        """
+        Return number of hot items in dataset.
+        Items with internal ids from `0` to `n_hot_items - 1` are hot (they re present in interactions).
```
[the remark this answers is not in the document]
feldlime: fixed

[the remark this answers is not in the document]
feldlime: fixed above (duplicate)

rectools/model_selection/last_n_split.py
feldlime: fixed

rectools/model_selection/last_n_split.py
feldlime: fixed

feldlime: fixed

rectools/model_selection/last_n_split.py
feldlime: fixed

rectools/model_selection/random_split.py
feldlime: fixed

rectools/model_selection/random_split.py
```diff
@@ -45,8 +45,10 @@ class RandomSplitter(Splitter):
...
+        WARNING: both cold and warm items will be excluded from test.
     filter_already_seen : bool, default ``True``
         If `True`, pairs (user, item) that are in train will be excluded from test.
```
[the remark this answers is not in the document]
feldlime: fixed

rectools/model_selection/random_split.py
feldlime: fixed

rectools/model_selection/time_split.py
feldlime: fixed

rectools/model_selection/time_split.py
```diff
@@ -57,8 +57,10 @@ class TimeRangeSplitter(Splitter):
...
+        WARNING: both cold and warm users will be excluded from test.
     filter_cold_items : bool, default ``True``
         If `True`, items that not in train will be excluded from test.
+        WARNING: both cold and warm items will be excluded from test.
     filter_already_seen : bool, default ``True``
         If ``True``, pairs (user, item) that are in train will be excluded from test.
```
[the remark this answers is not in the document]
feldlime: fixed

rectools/model_selection/cross_validate.py
feldlime: renamed to prefer_warm_inference_over_cold

rectools/models/base.py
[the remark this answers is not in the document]
feldlime: > So there is absolutely no way that a cold item could be recommended to itself.

Theoretically there may be an exception from this rule. E.g. if you ask Popular model for 10 recos for cold items and you give it 10 items in `items_to_recommend` list, then it can return you all of them, even if they are cold. However we doesn't support it now.

I agree with the rest. Simplified the code

rectools/models/base.py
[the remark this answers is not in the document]
feldlime: > Accept Union[InternalIds, List[InternalId]] in functions where we truly need to

Usually we don't need to accept such type. Otherwise it's easier to return a `list` instead of `np.array`.

> I'll ask for your comments

In current PR?

rectools/models/base.py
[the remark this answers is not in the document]
blondered: Also can we still drop pandas for filtering?
For example, `_filter_item_itself_from_i2i_reco` will just do the filtering with numpy. And we can make the `head(k)` part in `_make_reco_table` with provided `check_k` argument.

rectools/models/base.py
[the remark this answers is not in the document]
blondered: That's exactly what I was thinking. Usually we don't  accept the union of list and np.array. So we can clarify things if we fix InternalIds as np.array.

> In current PR?
Yes, please. For ModelBase private recommend methods that should be overwritten by children classes. Which of them are purely internal and which can proceed both internal+external target ids.

rectools/models/base.py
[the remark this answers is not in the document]
feldlime: But in this case we will execute `.groupby().head()` for all reco, even for cold where we don't need it

rectools/models/base.py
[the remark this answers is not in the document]
feldlime: Ok, let me think about it later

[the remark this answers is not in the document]
feldlime: Discussed separately.  
I added 3 more types `...Array` and used them

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [0.6.0] - Unreleased
+
+### Added 
+- Warm and cold users/items support to models ([#77](https://github.com/MobileTeleSystems/RecTools/pull/77))
```
blondered, line 12:
```suggestion
- Warm and cold users/items support in `ModelBase` ([#77](https://github.com/MobileTeleSystems/RecTools/pull/77))
```

feldlime: done

blondered (approved):

PR #80, opened 2023-12-28
Feature/visuals

blondered:
## Description

Visual analysis app for recommendations:
- demo: https://recsysart.ru/voila/
- supports comparing different models and requests
- supports u2i and i2i reco cases
- supports predefined requests
- supports `formatters` to create any html code from dataframe content

Todo in next PR:
- support for random requests
- support for saving and loading data for the app (since widgets disappear from ipynb on closure)
- examples (adding before release)
- new extra dependencies description in readme

## Type of change

- [ ] Optimization

## How Has This Been Tested?


rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+    """
+    Storage and processing of data for `VisualApp` widgets. This class is not meant to be used
+    directly. Use `VisualApp` class instead
+    """
+
+    def __init__(
+        self,
+        recos: tp.Dict[tp.Hashable, pd.DataFrame],
+        item_data: pd.DataFrame,
+        requests_dict: tp.Dict[tp.Hashable, tp.Hashable],
```
feldlime, line 21: you don't use `_dict` part in `recos` argument, but use here

I think we can remove it

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+        self.interactions: pd.DataFrame = interactions
+
+        if not self.requests_dict:
+            raise ValueError("`requests_dict` is empty")
+        self.processed_interactions = self._process_interactions()
+        self.processed_recos = self._process_recos()
+
+    def _process_interactions(self) -> tp.Dict[tp.Hashable, pd.DataFrame]:
+        prepared_interactions = {}
+        for request_name, request_id in self.requests_dict.items():
```
feldlime, line 53: `request_id` -> `request_value`? or just `value`?

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+            prepared_interactions[request_name] = (
+                self.interactions[self.interactions[self.request_colname] == request_id]
+                .merge(self.item_data, how="left", on=Columns.Item)
+                .drop(columns=[self.request_colname])
+            )
+        return prepared_interactions
+
+    def _process_recos(self) -> tp.Dict[tp.Hashable, tp.Dict[tp.Hashable, pd.DataFrame]]:
```
feldlime, line 61: Let's add type like `Tables` or `TablesDict` if you prefer for `tp.Dict[tp.Hashable, pd.DataFrame]`

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+        prepared_recos = {}
+        for model_name, full_recos in self.recos.items():
```
feldlime, line 63: How can be `full_recos` a part of just `recos`?

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+            model_recos = {}
+            for request_name, request_id in self.requests_dict.items():
```
feldlime, line 65: Don't you think it can be more effective to
1. Filter all request ids using `isin` from `full_recos`
2. merge
3. separate
?

Same for interactions

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+    recos : tp.Dict[tp.Hashable, pd.DataFrame]
+        Recommendations from different models in a form of a dict. Model names are supposed to be
+        dict keys. Recommendations from models are supposed to be in form of pandas DataFrames with
+        columns:
+            - `Columns.Item` - recommended item id
+            - `Columns.User` - user id (only for u2i)
+            - `Columns.TargetItem` - target item id (only for i2i)
+            - Any other columns that you wish to display in widgets (e.g. rank or score)
+        The original order of the rows will be preserved. Keep in mind to sort the rows correctly
+        before visualizing. The most intuitive wy is to sort by rank in ascending order.
```
feldlime, line 116:
```suggestion
        before visualizing. The most intuitive way is to sort by rank in ascending order.
```

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+    item_data : pd.DataFrame
+        Data for items that is used for visualisation in both interactions and recos widgets.
+        Supposed to be in form of a pandas DataFrame with columns:
+            - `Columns.Item` - item id
+            - Any other columns with item data (e.g. name, category, popularity, image link)
+    requests_dict : tp.Dict[tp.Hashable, tp.Hashable]
+        Predefined requests that will be displayed in widgets. For u2i case specific users are
+        considered as requests. For i2i case - specific items. Their names must be specified as keys
+        of the dict and ids as values of the dict.
+    is_u2i : bool, optional
```
feldlime, line 126: Please add default value to the line where argument is defined

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+        Display widgets right after initialization, by default True
+    formatters : tp.Optional[tp.Dict[str, tp.Callable]], optional
+        Formatter functions to apply to columns elements in the sections of interactions and recos.
+        Keys of the dict must be columns names (item_data, interactions and recos columns can be
+        specified here). Values bust be functions that will be applied to corresponding columns
+        elements. The result of each function must be a unicode string that represents html code.
+        Formatters can be used to format text, create links and display images with html.
+        By default None.
+    rows_limit : int, optional
+        Maximum number of rows to display in the sections of interactions and recos, by default 20
```
feldlime, line 147:
```suggestion
        Maximum number of rows to display in the sections of interactions and recos, by default 20.
```

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+    def __init__(
+        self,
+        recos: tp.Dict[tp.Hashable, pd.DataFrame],
+        item_data: pd.DataFrame,
+        requests_dict: tp.Dict[tp.Hashable, tp.Hashable],
+        is_u2i: bool = True,
+        interactions: tp.Optional[pd.DataFrame] = None,
+        auto_display: bool = True,
+        formatters: tp.Optional[tp.Dict[str, tp.Callable]] = None,
+        rows_limit: int = 20,
```
feldlime, line 184: Don't you think 20 is too many?

tests/visuals/__init__.py
```diff
@@ -0,0 +1,13 @@
+#  Copyright 2022 MTS (Mobile Telesystems)
```
feldlime, line 1: Do you know is it correct that we're still using 2022?

feldlime: And btw probably we should add it to every file?

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+        """Display model_name"""
+        display(widgets.HTML(value=f"Model name: {model_name}"))
+
+    def display(self) -> None:
+        """Display full VisualApp widget"""
+        request_name_selection = widgets.ToggleButtons(
+            options=self.data_storage.request_names,
+            description=f"Request {self.data_storage.request_colname}:",
+            disabled=False,
+            button_style="warning",
```
feldlime, line 243: I understand that you just need yellow color, but probably there is way to apply it without using warning style?

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+    Storage and processing of data for `VisualApp` widgets. This class is not meant to be used
+    directly. Use `VisualApp` class instead
+    """
+
+    def __init__(
+        self,
+        recos: tp.Dict[tp.Hashable, pd.DataFrame],
+        item_data: pd.DataFrame,
+        requests_dict: tp.Dict[tp.Hashable, tp.Hashable],
+        is_u2i: bool = True,
```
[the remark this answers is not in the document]
blondered: I see your point. For example I can think of simple visualisation of predefined item lists (like hand-made recos and popular recos for cold scenarios) without the need for interactions history at all. 
But if we think that this is going to be an option, should we move all of the use cases to different classes? Because each option has very strict demands for the input data. And it is going to be very confusing to try to figure out necessary parameters for each case.
Like: pass interactions for u2i but don't pass for i2i and cold. And so on

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+import numpy as np
+import pandas as pd
+from IPython.display import display
+
+from rectools import Columns
+
+
+class AppDataStorage:
+    """
+    Storage and processing of data for `VisualApp` widgets. This class is not meant to be used
```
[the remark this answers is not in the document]
blondered: I believe final decision should be made together with the functionality to save and upload data for visualisation. And I want to add it in the next PR. We must add another PR before releasing this app anyway.
So can we leave it as it is for now?

rectools/visuals/visual_app.py
blondered: It's ok, I'm using it at 20 and it works fine. If user has 20 rows in recos df, he knows why he needs all of them :)

rectools/visuals/visual_app.py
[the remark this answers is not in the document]
feldlime: ok

rectools/visuals/visual_app.py
feldlime: ok

rectools/visuals/visual_app.py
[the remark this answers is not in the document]
feldlime: Yes, probably having a separate class for each case is a good idea

rectools/visuals/visual_app.py
blondered: I wish there was a better way. But it is only possible through some hacks like these:
- https://stackoverflow.com/questions/68643117/python-ipywidgets-togglebutton-style-color
- https://stackoverflow.com/questions/72504234/styling-ipywidgets-accordion-panels-individually

Until ipywidgets add support for `button_color` property in `ToggleButtons.style` we are stuck with predefined colors

tests/visuals/__init__.py
blondered: I think it works like "since 2022". But I've fixed for newly added file

rectools/visuals/visual_app.py
blondered: fixed

rectools/visuals/visual_app.py
blondered: fixed

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+                        full_recos[full_recos[self.request_colname] == request_id],
+                        how="right", 
+                        on="item_id",
+                        suffixes=["_item", "_recos"])
+                    .drop(columns=[self.request_colname])
+                )
+            prepared_recos[model_name] = model_recos
+        return prepared_recos
+
+    def _check_columns_present_in_recos(self, recos: tp.Dict[tp.Hashable, pd.DataFrame]) -> None:
```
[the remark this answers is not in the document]
blondered: fixed

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+        self.recos = recos
+
+        if Columns.Item not in item_data:
+            raise KeyError(f"Missed {Columns.Item} column in item_data")
+        self.item_data = item_data
+
+        if interactions is None and is_u2i:
+            raise ValueError("For u2i reco you must specify interactions")
+        if interactions is not None and not is_u2i:
+            raise ValueError("For i2i reco you shouldn't specify interactions")
```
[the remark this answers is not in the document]
blondered: fixed

rectools/visuals/visual_app.py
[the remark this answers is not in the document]
blondered: Check it out. Not sure about ABC and abstract method decorator here. Should we add it? Also I decided to change AppDataStorage to a private class after all. Probably VisualAppBase should also be private?

rectools/visuals/visual_app.py
```diff
@@ -0,0 +1,273 @@
...
+        if not is_u2i:
+            interactions = self._prepare_interactions_for_i2i()
+        self.interactions: pd.DataFrame = interactions
+
+        if not self.requests_dict:
+            raise ValueError("`requests_dict` is empty")
+        self.processed_interactions = self._process_interactions()
```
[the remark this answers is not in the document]
blondered: no ideas for now. `split_to_requests_and_fill_with_item_data` doesn't seem any better :)
`prepare_interactions_for_visual_app`?

rectools/visuals/visual_app.py
blondered: fixed

rectools/visuals/visual_app.py
blondered: it's actually the id (user or item). and it is important, so definitely `request_id`

rectools/visuals/visual_app.py
blondered: fixed

rectools/visuals/visual_app.py
blondered: fixed. but it also went to the docstrings for users. is it ok?

rectools/visuals/visual_app.py
blondered: fixed

rectools/visuals/visual_app.py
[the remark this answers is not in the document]
feldlime: Don't use ABC and decorators please, use `NotImplemented` / `NotImplementedError` instead if you really need it.

I really don't see any reason to make `VisualAppBase` private. We have base classes for models, metrics, splitters and all of them are public.

rectools/visuals/visual_app.py
feldlime: ok

But it's not obvious while reading the code

rectools/visuals/visual_app.py
feldlime: It's a point to think about. It's ok for now, maybe we will change it later

tests/visuals/__init__.py
feldlime: Can you please ask smb about it? I'm sure we have same problems for all new files

rectools/visuals/visual_app.py
feldlime: Hacks don't look too bad

Let's keep it as is, but add some comments please with this links probably

rectools/visuals/visual_app.py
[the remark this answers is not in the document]
blondered: Ok. This thread is ready for review. Let's get the final version of VisualApp classes in this PR. And work on final version of DataStorage in the next PR

rectools/visuals/visual_app.py
blondered: They are hard to support and impossible to check during linters and tests. So I really don't want to add them. I've added explicit comment to code.

rectools/visuals/visual_app.py
blondered: added comment

rectools/visuals/visual_app.py
blondered: ok

rectools/visuals/visual_app.py
blondered: Yes, it gives small boost for large dataframes. Fixed

tests/visuals/__init__.py
blondered: Ok, waiting for info here

rectools/visuals/visual_app.py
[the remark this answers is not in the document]
blondered: `group` and `grouped` is my final :)

feldlime (approved):

PR #107, opened 2024-02-26
Feature/ease

In48semenov:
## Description

Implementation of the [ease](https://arxiv.org/pdf/1905.03375.pdf) model

For u2i recommendations I use ImplicitRanker. It's in the [vector.py](https://github.com/MobileTeleSystems/RecTools/blob/main/rectools/models/vector.py#L47), but the EASE model is inherited from [ModelBase](https://github.com/MobileTeleSystems/RecTools/blob/main/rectools/models/base.py#L30). In addition, ImplicitRanker accepts [subjects_factors as input, like np.ndarray](https://github.com/MobileTeleSystems/RecTools/blob/main/rectools/models/vector.py#L61), but I pass scipy.sparse.csr_matrix as input, which speeds up the inference a little.
I propose to move ImplicitRanker into a separate module and add types of accepted objects.

## Type of change

- [ ] Optimization

## How Has This Been Tested?

### DIFF
--- a/rectools/models/ease.py
+++ b/rectools/models/ease.py
@@ -101,8 +101,8 @@ class EASE(ModelBase):
             viewed_ids = get_viewed_item_ids(user_items, user_id)  # sorted
         else:
             viewed_ids = np.array([], dtype=int)
-
-        scores = user_items[user_id].dot(self.weight).getA1()
+        
+        scores = (user_items[user_id] @ self.weight).getA1()
         reco_ids, reco_scores = recommend_from_scores(
             scores=scores, k=k, sorted_blacklist=viewed_ids, sorted_whitelist=sorted_item_ids
         )
--- a/rectools/models/ease.py
+++ b/rectools/models/ease.py
@@ -24,8 +24,8 @@ from rectools import InternalIds
 from rectools.dataset import Dataset
 
 from .base import ModelBase, Scores
+from .utils import recommend_from_scores
 from .vector import Distance, ImplicitRanker
-from .utils import get_viewed_item_ids, recommend_from_scores
 
 
 class EASEModel(ModelBase):
--- a/rectools/models/ease.py
+++ b/rectools/models/ease.py
@@ -28,7 +28,7 @@ from .vector import Distance, ImplicitRanker
 from .utils import get_viewed_item_ids, recommend_from_scores
 
 
-class EASE(ModelBase):
+class EASEModel(ModelBase):
     """
     Embarrassingly Shallow Autoencoders for Sparse Data model.
 
@@ -51,7 +51,7 @@ class EASE(ModelBase):
     ):
         super().__init__(verbose=verbose)
         self.regularization = regularization
-        self.weight: np.matrix
+        self.weight: np.ndarray
 
     def _fit(self, dataset: Dataset) -> None:  # type: ignore
         ui_csr = dataset.get_user_item_matrix(include_weights=True)
@@ -62,7 +62,7 @@ class EASE(ModelBase):
 
         gram_matrix_inv = np.linalg.inv(gram_matrix)
 
-        self.weight = gram_matrix / (-np.diag(gram_matrix_inv))
+        self.weight = np.array(gram_matrix / (-np.diag(gram_matrix_inv)))
         np.fill_diagonal(self.weight, 0.0)
 
     def _recommend_u2i(
@@ -78,7 +78,7 @@ class EASE(ModelBase):
         ranker = ImplicitRanker(
             distance=self.u2i_dist,
             subjects_factors=user_items,
-            objects_factors=np.array(self.weight),
+            objects_factors=self.weight,
         )
         ui_csr_for_filter = user_items[user_ids] if filter_viewed else None
 
@@ -92,25 +92,6 @@ class EASE(ModelBase):
 
         return all_user_ids, all_reco_ids, all_scores
 
-    def _recommend_for_user(
-        self,
-        user_id: int,
-        user_items: sparse.csr_matrix,
-        k: int,
-        filter_viewed: bool,
-        sorted_item_ids: tp.Optional[np.ndarray],
-    ) -> tp.Tuple[InternalIds, Scores]:
-        if filter_viewed:
-            viewed_ids = get_viewed_item_ids(user_items, user_id)  # sorted
-        else:
-            viewed_ids = np.array([], dtype=int)
-        
-        scores = (user_items[user_id] @ self.weight).getA1()
-        reco_ids, reco_scores = recommend_from_scores(
-            scores=scores, k=k, sorted_blacklist=viewed_ids, sorted_whitelist=sorted_item_ids
-        )
-        return reco_ids, reco_scores
-
     def _recommend_i2i(
         self,
         target_ids: np.ndarray,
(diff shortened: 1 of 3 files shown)


rectools/models/ease.py
```diff
@@ -0,0 +1,134 @@
...
+            objects_factors=self.weight,
+        )
+        ui_csr_for_filter = user_items[user_ids] if filter_viewed else None
+
+        all_user_ids, all_reco_ids, all_scores = ranker.rank(
+            subject_ids=user_ids,
+            k=k,
+            filter_pairs_csr=ui_csr_for_filter,
+            sorted_object_whitelist=sorted_item_ids_to_recommend,
+            num_threads=0,
```
blondered, line 90: num_threads should be passed as a parameter for EASE model as in LightFM:
https://github.com/MobileTeleSystems/RecTools/blob/c5ca01a8def82a7d32b947bb386b96d98719e50a/rectools/models/lightfm.py#L44
here we should call it with defined values

rectools/models/ease.py
```diff
@@ -0,0 +1,134 @@
...
+        sorted_item_ids_to_recommend: tp.Optional[np.ndarray],
+    ) -> tp.Tuple[InternalIds, InternalIds, Scores]:
+        similarity = self.weight
+        if sorted_item_ids_to_recommend is not None:
+            similarity = similarity[:, sorted_item_ids_to_recommend]
+
+        all_target_ids = []
+        all_reco_ids: tp.List[np.ndarray] = []
+        all_scores: tp.List[np.ndarray] = []
+        for target_id in tqdm(target_ids, disable=self.verbose == 0):
```
blondered, line 109: Cycles of this type are very slow for inference. We do this only when we have no other option. But here you have a very nice filled item-item similarity matrix in np.array format.

If we have similarity scores array for one item and need recommendations for it, we can do:
```
unsorted_reco_positions = scores.argpartition(-k)[-k:]
unsorted_reco_scores = scores[unsorted_reco_positions]
order = unsorted_reco_scores.argsort()[::-1]
reco = unsorted_reco_positions[order]
scores = unsorted_reco_scores[order]
```

But in EASE recommend we have scores as a matrix and need to process recommendations for a couple of items. We can do smth like:

```
unsorted_reco_positions = scores.argpartition(-k, axis=1)[:, -k:]
unsorted_reco_scores = np.take_along_axis(scores, unsorted_reco_positions, axis=1)
reco = np.take_along_axis(
            unsorted_reco_positions, unsorted_reco_scores.argsort()[:, ::-1], axis=1
        )
```

blondered (changes requested):

rectools/models/ease.py
blondered: We should probably add batches processing for this logic.

Another way is to use ImplicitRanker again. Just feed identity matrix as subject factors. It could be faster this way due to implicit library realisation.

In48semenov: It might be worth changing the `recommend_from_scores` function so that it works with scores with dimensions `(1, n_items)` rather than `(n_items,)`?

rectools/models/rank.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def rank(
+        self,
+        subject_ids: InternalIds,
+        k: int,
+        filter_pairs_csr: tp.Optional[sparse.csr_matrix] = None,  # subect-object interactions, relevant for u2i case
+        sorted_object_whitelist: tp.Optional[np.ndarray] = None,
+        num_threads: int = 0,
+    ) -> tp.Tuple[InternalIds, InternalIds, Scores]:
+        """Proceed inference using implicit library matrix factorization topk cpu method"""
```
blondered, line 136: Since we're making this a common method for a few model types, please write a full docstring for it and describe arguments.

rectools/models/rank.py
```diff
@@ -0,0 +1,181 @@
...
+    """Distance metric"""
+
+    DOT = 1  # Bigger value means closer vectors
+    COSINE = 2  # Bigger value means closer vectors
+    EUCLIDEAN = 3  # Smaller value means closer vectors
+
+
+class ImplicitRanker:
+    """
+    Ranker model which uses implicit library matrix factorization topk method
```
blondered, line 39: Let's expand here: "This ranker is suitable for the following cases of scores calculation: `subject_embeddings.dot(objects_embeddings)` and `subject_interactions.dot(item-item-similarities)`"

rectools/models/rank.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Parameters
+    ----------
+    distance : Distance
+        Distance metric.
+    subjects_factors : np.ndarray | sparse.csr_matrix
+        Array of subject embeddings, shape (n_subjects, n_factors).
```
blondered, line 46:
```suggestion
        Array of subjects embeddings, shape (n_subjects, n_factors). For item-item similarity models subjects vectors from ui_csr are viewed as factors.
```

rectools/models/rank.py
```diff
@@ -0,0 +1,181 @@
...
+    objects_factors : np.ndarray
+        Array with embeddings of all objects, shape (n_objects, n_factors).
```
blondered, line 48:
```suggestion
        Array with embeddings of all objects, shape (n_objects, n_factors). For item-item similarity models item similarity vectors are viewed as factors.
```

blondered (changes requested):

rectools/models/ease.py
```diff
@@ -0,0 +1,121 @@
...
+        similarity = self.weight[target_ids]
+        if sorted_item_ids_to_recommend is not None:
+            similarity = similarity[:, sorted_item_ids_to_recommend]
+
+        n_reco = min(k, similarity.shape[1])
+        unsorted_reco_positions = similarity.argpartition(-n_reco, axis=1)[:, -n_reco:]
+        unsorted_reco_scores = np.take_along_axis(similarity, unsorted_reco_positions, axis=1)
+
+        all_reco_scores = np.take_along_axis(unsorted_reco_scores, unsorted_reco_scores.argsort()[:, ::-1], axis=1)
+        all_reco_ids = np.take_along_axis(unsorted_reco_positions, unsorted_reco_scores.argsort()[:, ::-1], axis=1)
```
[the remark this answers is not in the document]
In48semenov: Yes, 'all_reco_ids' also have the same meaning as in 1d case

rectools/models/rank.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def rank(
+        self,
+        subject_ids: InternalIds,
+        k: int,
+        filter_pairs_csr: tp.Optional[sparse.csr_matrix] = None,
+        sorted_object_whitelist: tp.Optional[np.ndarray] = None,
+        num_threads: int = 0,
+    ) -> tp.Tuple[InternalIds, InternalIds, Scores]:
+        """Rank objects for proceed inference with using implicit library topk cpu method.
```
blondered, line 142:
```suggestion
        """Rank objects to proceed inference using implicit library topk cpu method.
```

rectools/models/rank.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Parameters
+        ----------
+        subject_ids : csr_matrix
+            Array of ids to recommend for.
+        k : int
+            Derived number of recommendations for every subject ids.
```
blondered, line 149:
```suggestion
            Derived number of recommendations for every subject id.
```

rectools/models/rank.py
```diff
@@ -0,0 +1,209 @@
...
+        filter_pairs_csr : sparse.csr_matrix, optional, default ``None``
+            Subject-object interactions.
```
blondered, line 151:
```suggestion
            Subject-object interactions that should be filtered from recommendations.
```

rectools/models/ease.py
```diff
@@ -0,0 +1,127 @@
...
+from .rank import Distance, ImplicitRanker
+
+
+class EASEModel(ModelBase):
+    """
+    Embarrassingly Shallow Autoencoders for Sparse Data model.
+
+    See https://arxiv.org/abs/1905.03375.
+
+    Please note that this algorithm requires a lot of RAM during fit method.
```
blondered, line 35:
```suggestion
    Please note that this algorithm requires a lot of RAM during `fit` method.
```

rectools/models/ease.py
```diff
@@ -0,0 +1,127 @@
...
+    Reasonable amount of interactions is about 20m.
+
+    Parameters
+    ----------
+    regularization : float
+        The regularization factor of the weights.
+    verbose : int, default 0
+        Degree of verbose output. If 0, no output will be provided.
+    num_threads: int, default 1
+        Number of threads used for recommend method.
```
blondered, line 47:
```suggestion
        Number of threads used for `recommend` method.
```

blondered (commented):
We are ready to merge this, just a few small last changes please :)

blondered (approved):

Issue #90, opened 2024-02-21
Users parameter for cross-validate


blondered:
### Feature Description

`users` parameter in `cross-validate` will accept the list of external user ids to run all experiments only on these users.

### Why this feature?

This will help to solve a common task to calculate metrics only on specific subset of users

nishantguvvada:
Hi @blondered, I hope it's okay for me to go over it. I've created a pull request in response to the issue: https://github.com/MobileTeleSystems/RecTools/pull/112

PR #112, opened 2024-03-01
Added users parameter in cross-validate

nishantguvvada:
## Description

Added users parameter in cross-validate.py to the request : https://github.com/MobileTeleSystems/RecTools/issues/90

 This will help to solve a common task to calculate metrics only on specific subset of users

## Type of change

- [ ] Optimization

## How Has This Been Tested?

### DIFF
--- a/rectools/model_selection/cross_validate.py
+++ b/rectools/model_selection/cross_validate.py
@@ -48,6 +48,7 @@ def cross_validate(  # pylint: disable=too-many-locals
     models: tp.Dict[str, ModelBase],
     k: int,
     filter_viewed: bool,
+    users: tp.Optional[ExternalIds] = None,
     items_to_recommend: tp.Optional[ExternalIds] = None,
 ) -> tp.Dict[str, tp.Any]:
     """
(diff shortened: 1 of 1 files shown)


rectools/model_selection/cross_validate.py
```diff
@@ -48,6 +48,7 @@ def cross_validate(  # pylint: disable=too-many-locals
     models: tp.Dict[str, ModelBase],
     k: int,
     filter_viewed: bool,
+    users: tp.Optional[ExternalIds] = None,
```
blondered, line 51: Hi! Thank you for contribution!

Could you please add the following to your PR:
1. Make `users` parameter as the last last one
2. Add profound testing in `tests/model_selection/test_cross_validate.py`. Please ensure that math is correctly checked there. We believe that tests are really important
3. Add your change in CHANGELOG.md

rectools/model_selection/cross_validate.py
```diff
@@ -115,6 +116,8 @@ def cross_validate(  # pylint: disable=too-many-locals
         fold_dataset = _gen_2x_internal_ids_dataset(interactions_df_train, dataset.user_features, dataset.item_features)
 
         interactions_df_test = interactions.df.iloc[test_ids]  # 1x internal
+        if users is not None:
+            interactions_df_test = interactions_df_test[interactions_df_test[Columns.User].isin(users)]
```
blondered, line 120: Here we already moved to internal ids for faster calculations. So it is necessary to convert external ids to internal. You can use `dataset.user_id_map` for this

blondered (changes requested):

rectools/model_selection/cross_validate.py
nishantguvvada: Thanks for the feedback, I'm on it.

rectools/model_selection/cross_validate.py
nishantguvvada: I've added the convert_to_internal method here

Issue #119, opened 2024-04-04
Cold users/items support for non-personalized models


feldlime:
Add the ability to recommend for cold users and items to models:
- `PopularModel`
- `PopularInCategoryModel`
- `RandomModel`

PR #120, opened 2024-04-04
Cold users/items support in non-personilized models

feldlime:
## Description

- Supported cold recommendations in non-personalized models (random, popular and popular-in-category)
- Added check for incorrect types of cold targets
- Changed the logic of choosing random sampler for `RandomModel` and increased the sampling speed
- Changed the logic of `RandomModel`: now the recommendations are different for repeated calls of recommend methods

Closes https://github.com/MobileTeleSystems/RecTools/issues/119

## Type of change

- [ ] Optimization

## How Has This Been Tested?


rectools/models/random.py
```diff
@@ -118,3 +118,27 @@ def _recommend_i2i(
         sorted_item_ids_to_recommend: tp.Optional[InternalIdsArray],
     ) -> tp.Tuple[InternalIds, InternalIds, Scores]:
         return self._recommend_u2i(target_ids, dataset, k, False, sorted_item_ids_to_recommend)
+
+    def _get_filtered_item_ids(self, sorted_item_ids_to_recommend: tp.Optional[InternalIdsArray]) -> InternalIdsArray:
+        if sorted_item_ids_to_recommend is not None:
+            item_ids = np.unique(sorted_item_ids_to_recommend)
```
blondered, line 124: `sorted_item_ids_to_recommend` are already after `np.unique`. Why do we do it here?

rectools/models/random.py
```diff
@@ -22,17 +22,26 @@
...
+        self.values = values
+        self.values_list = list(values)  # for random.sample
+        self.python_gen = random.Random(random_state)
+        self.np_gen = np.random.default_rng(random_state)
+
+    def sample(self, n: int) -> np.ndarray:
+        if n < 25:  # Empiric value, for optimization
+            sampled = np.asarray(self.python_gen.sample(self.values_list, n))
+        else:
+            sampled = self.np_gen.choice(self.values, n, replace=False)  # pragma: no cover
```
blondered, line 43: Maybe add test here? Happy path at least?

blondered, line 43: xMaybe add test here? Happy path at least?

rectools/models/random.py
```diff
@@ -71,18 +83,11 @@ def _recommend_u2i(
...
-            item_ids = np.unique(sorted_item_ids_to_recommend)
-        else:
-            item_ids = self.all_item_ids
-
-        item_indices = list(range(item_ids.size))  # for random.sample
-
-        np.random.seed(self.random_state)
-        random.seed(self.random_state, version=2)
+        item_ids = self._get_filtered_item_ids(sorted_item_ids_to_recommend)
+        sampler = _RandomSampler(item_ids, self.random_state)
```
blondered, line 87: We will have the same recommendations for multiple `recommend` calls. Is this our desired behaviour? I'm not sure.

rectools/models/random.py
feldlime: duplicate?

rectools/models/random.py
feldlime: It's a good point to think.

Actually we had the same behaviour before, so I haven't changed anything. 
But probably it'd be really better to change this.

rectools/models/random.py
feldlime: Fixed

rectools/models/random.py
feldlime: fixed

rectools/models/random.py
feldlime: Added

rectools/models/random.py
blondered: Probably should add this to Changelog

rectools/models/random.py
```diff
@@ -118,3 +126,27 @@ def _recommend_i2i(
         sorted_item_ids_to_recommend: tp.Optional[InternalIdsArray],
     ) -> tp.Tuple[InternalIds, InternalIds, Scores]:
         return self._recommend_u2i(target_ids, dataset, k, False, sorted_item_ids_to_recommend)
+
+    def _get_filtered_item_ids(self, sorted_item_ids_to_recommend: tp.Optional[InternalIdsArray]) -> InternalIdsArray:
```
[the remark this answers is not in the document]
feldlime: removed

blondered (approved):

PR #130, opened 2024-05-14
Feature/random state to pure svd

feldlime:
## Description

- Added `random_state`, `maxiter` and `tol` parameters to the PureSVD model
- Refactored `Intersection` metric
- Fixed `coverage` part in the `test` workflow
- Fixed coverage of the code 

## Type of change

- [ ] Optimization

## How Has This Been Tested?


rectools/metrics/intersection.py
```diff
@@ -71,24 +70,15 @@ def calc_per_user(self, reco: pd.DataFrame, ref_reco: pd.DataFrame) -> pd.Series
...
             Values of metric (index - user id, values - metric value for every user).
         """
-        self._check(reco)
-        assert set(ref_reco.columns) >= {Columns.User, Columns.Item, Columns.Rank}
+        self._check(reco, ref_reco=ref_reco)
 
         if ref_reco.shape[0] == 0:
             return pd.Series(index=pd.Series(name=Columns.User, dtype=int), dtype=np.float64)
 
-        if ref_reco is reco:
```
blondered, line 80: Are you sure we should drop this? In `cross_validate` we will get boost to speed without actually calculating recall because we are passing the same object for both `reco` and `ref_reco`

tests/models/test_pure_svd.py
```diff
@@ -194,8 +194,10 @@ def test_i2i(
             actual,
         )
 
-    def test_second_fit_refits_model(self, dataset: Dataset) -> None:
-        model = PureSVDModel(factors=3)
+    @pytest.mark.parametrize("tol, random_state", ((0, None), (0, 42), (0.3, 42)))
+    def test_second_fit_refits_model(self, dataset: Dataset, tol: float, random_state: tp.Optional[int]) -> None:
+        # If `tol` is greater than 0, and `random_state` is None, the results can be different.
```
blondered, line 199: What actually happens in this case?

rectools/metrics/intersection.py
feldlime: I totally forgot about the `cross_validate`, thanks. Refactored it as well

Returned this optimisation

tests/models/test_pure_svd.py
feldlime: You mean why the results are different?

blondered: What do you mean by different results? Second `fit` doesn't refit model? Or we just can't check it because random state is not set and we get different results each fit.

feldlime: If random state is not set we get different results each fit (it's not directly connected with refit).  It worked before because results are the same even without random state if `tol == 0`

blondered (approved):

Issue #137, opened 2024-05-14
Extended RecSys baselines tutorial


blondered:
### Feature Description

Extended tutorial for all of RecSys baselines: theory and practice

### Why this feature?

Accumulate expertise together with code

PR #139, opened 2024-05-15
Extended tutorial for baselines

spirinamayya:
## Description

Added extended tutorial for baselines
Closes https://github.com/MobileTeleSystems/RecTools/issues/137


examples/8_baseline_extended_tutorial.ipynb
feldlime, line 1: 1. Please add this example to `docs/source/tutorials.rst`
2. Probably it's better to use the English version of the dataset, @blondered WDYT?
3. Let's give a short explanation what Kion dataset is. Just a couple of sentences. And maybe `interactions/users/items.head(2) + ...shape` for better understanding.
4. Let's also show `user_features` and `item_features` shape + head, otherwise it's too hard to understand.
5. For users selected as test users it's probably makes sense to show interactions and features (or prove the absence). WDYT?
6. iALS: since you give a formula and comments to it, let's write what alpha and r_ui are.
7. We don't apply linters to notebooks yet, so please follow the common code style: closing bracket should be on the new line (for multiline cases), and commas should be on each line.
8. Not sure if it makes sense to show the titles of recommended movies since we don't show the history of user interactions. WDYT?
9. LightFM: it is worth showing in text and/or in code that the number of epochs may (and should) be specified.
10. LightFM: for a newcomer (and not only for them) the similar -220... scores for the 1st and 2nd users may look pretty weird especially given that one of them has features and especially in contrast with the "normal" scores for the last user. Let's give a comment maybe.
11. SVD: you're using `dataset_no_features` here since SVD doesn't use features. But I'm afraid it may give a false impression that it's impossible to pass a dataset with features to this model, and that our user should always create 2 different datasets if they want to test e.g. LightFM and SVD. Let's maybe use the same dataset (with features) everywhere? Your model descriptions are quite extensive to understand that model doesn't use features. WDYT?

feldlime:
Don't pay attention to the number of comments, I really like your notebook, it looks fabulous! I can't even imagine how much time it took to write such an extensive model descriptions and to draw such a nice pictures, and I really appreciate that. Thank you!

examples/8_baseline_extended_tutorial.ipynb
blondered: 1. +
2. We're not showing any russian in the tutorial but it is a good idea for users tht want to play around with the notebook. Agreed. I think it will be enough to change the link to the `data_en.zip`. But full notebook should be run of course.
3. Let's add: "We are using KION dataset for this tutorial. The data was gathered from the users of MTS KION video streaming platform and includes user-item interactions together with user and item features. [More info on the dataset](https://github.com/irsafilo/KION_DATASET)"
4. +
5. +
6. Alpha is fully explained in the 'model application' since it is not always called 'alpha' for all iALS implementations (I think). What exactly do you mean by r_ui here? Not sure it is necessary
7. + (but right now I don't see where the problems are
8. Titles need to be here, otherwise it's too boring :)
9. Could you right the final text? I think it sould be in the model application section as one of the points for practical advices.
10. Also please write the final text. We're getting deep here. 
11. Good point. We probably should drop `dataset_no_features` and user only one dataset everywhere. And add a point to every model with explanation. It's either "If we use dataset with features for fit and recommend, model just skips this info". Or "Use dataset with features for fit and recommend to add feature processing to the model"

blondered:
I must also emphasize, that we are still excepting (and will be excepting after merge) more links, explanations and practical advices from practitioners. **And we really appreciate everyone to join us sharing their expertise and accumulating it in one place.**

examples/8_baseline_extended_tutorial.ipynb
feldlime: 6. Here
image

7. E.g
image

8. Ok, let them be 

9. Just a bullet point in the list of model params like "`num_epochs` - number of model training epochs, 1 by default`
I think since we're naming the model params we should mention this one as well

10. I suggest the author to think about the final text here :) It can be just a comment like "We see score -220 for user x since it's a cold user and only item biasese were used for the score calculation....". Or, an alternative: we can make a short user descriptions like "hot user"/"warm user" in the beginning of the notebook, put them into a data frame and join to recos like movie titles. This would be even better I think.

blondered: 6. Ok, agreed. So we need  two points there after c_ui with alpha and r_ui explained
7. Ok. agreed
9. Ok, then it is needed for both iALS and LightFM. In the form of: "Specify `iterations` for number of model training epochs" for iALS just after regularization. And "Specify `epochs` for number of model training epochs" for LightFM after "pass LightFM model to the wrapper"

feldlime:
Please add one of the [keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) (I prefer `Closes`) before the issue link in the description

examples/8_baseline_extended_tutorial.ipynb
blondered: 10. Decided to keep this one for later. Not needed for this PR

docs/source/tutorials.rst
```diff
@@ -14,3 +14,4 @@ See tutorials here: https://github.com/MobileTeleSystems/RecTools/tree/main/exam
    examples/5_benchmark_iALS_with_features
    examples/6_benchmark_lightfm_inference
    examples/7_visualization
+   examples/8_baseline_extended_tutorial
```
blondered, line 17: This tutorial is a beautiful piece of work. So we have decided to move it to a proper place.
I'd like to ask you for a few last changes to make it shine more:
1. Let's create a subfolder in `examples` folder in repo with name `tutorials`
2. Let's move your tutorial to this subfolder
3. Let's change the name to `baselines_extended_tutorial.ipynb`
4. In the `tutorials.rst` keep only you tutorial, change the link in the header to tutorials subfolder in the repo
5. Create `examples.rst` where you move all necessary content which is now in `tutorials.rst`

This way we are distinguishing between examples (even long ones) and real theory & practice tutorials.

blondered: Let's also add `examples` to index.rst

docs/source/examples.rst
```diff
@@ -0,0 +1,16 @@
+Examples
```
blondered, line 1: Please add to CHANGELOG.MD under "Unreleased" "Added" header:
- Extended Theory&Practice RecSys baselines tutorial

docs/source/index.rst
```diff
@@ -100,3 +100,4 @@ The package also provides tools that allow to evaluate metrics as easy and as fa
    benchmarks
    faq
    support
+   examples
```
blondered, line 103: Let's move to a more reasonable place. Between API and Tutorials would be fine

blondered (approved):

PR #140, opened 2024-05-16
Feature/dual metrics widget

Gooogr:
## Description

Add a Plotly scatterplot widget for metrics visualization, enabling support for visualizing data per fold or their average values. 

## Type of change

- [ ] Optimization

## How Has This Been Tested?

Sorry, I can't figure out where I should place my changes in the changelog file. Should I create a new list of changes for the upcoming release 0.7.0?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -77,18 +77,19 @@ torch = {version = ">=1.6.0, <3.0.0", optional = true}
 pytorch-lightning = {version = ">=1.6.0, <3.0.0", optional = true}
 
 ipywidgets = {version = ">=7.7,<8.2", optional = true}
+plotly = "^5.22.0"
 
 
 [tool.poetry.extras]
 lightfm = ["lightfm"]
 nmslib = ["nmslib"]
 torch = ["torch", "pytorch-lightning"]
-visuals = ["ipywidgets"]
+visuals = ["ipywidgets", "plotly"]
 all = [
     "lightfm",
     "nmslib", 
     "torch", "pytorch-lightning",
-    "ipywidgets"
+    "ipywidgets", "plotly",
 ]
 
 
--- a/rectools/visuals/metrics_app.py
+++ b/rectools/visuals/metrics_app.py
@@ -1,0 +1,128 @@
+import typing as tp
+
+import ipywidgets as widgets
+import pandas as pd
+import plotly.express as px
+import plotly.graph_objects as go
+from IPython.display import display
+
+from rectools import Columns
+
+
+class MetricsApp:
+    """
+    Jupyter widgets app for metric visualization and comparison.
+    Do not create instances of this class directly. Use `MetricsApp.construct` method instead.
+    """
+
+    def __init__(
+        self,
+        df_metrics_data: pd.DataFrame,
+        chart_width: tp.Union[float, int] = 800,
+        chart_height: tp.Union[float, int] = 600,
+        show_legend: bool = False,
+        auto_display: bool = True,
+        color_discrete_sequence: tp.Optional[tp.List[str]] = None,
+    ):
+        self.metrics_data = df_metrics_data
+        self.chart_width = chart_width
+        self.chart_height = chart_height
+        self.show_legend = show_legend
+        self.auto_display = auto_display
+        self.color_discrete_sequence = (
+            px.colors.qualitative.Plotly if color_discrete_sequence is None else color_discrete_sequence
+        )
+
+        self._validate_input_data()
+        self._validate_input_patameters()
+        if self.auto_display:
+            self.display()
+
+    @classmethod
+    def construct(
+        cls,
+        df_metrics_data: pd.DataFrame,
+        chart_width: tp.Union[float, int] = 800,
+        chart_height: tp.Union[float, int] = 600,
+        show_legend: bool = False,
+        auto_display: bool = True,
+        color_discrete_sequence: tp.Optional[tp.List[str]] = None,
+    ) -> "MetricsApp":
+        """TODO: add docstring"""
+        return cls(df_metrics_data, chart_width, chart_height, show_legend, auto_display, color_discrete_sequence)
+
+    def _validate_input_data(self) -> None:
+        if not isinstance(self.metrics_data, pd.DataFrame):
+            raise ValueError("Incorrect input type. `metrics_data` should be a DataFrame")
+        if Columns.Split not in self.metrics_data.columns:
+            raise KeyError("Missing 'Split' column in `metrics_data` DataFrame")
+        if Columns.Model not in self.metrics_data.columns:
+            raise KeyError("Missing 'Model' column in `metrics_data` DataFrame")
+        if len(self.metrics_data.columns) < 3:
+            raise KeyError("`metrics_data` DataFrame assumed to have at least one metric column")
+
+    def _validate_input_patameters(self) -> None:
+        if self.chart_width < 10:
+            raise ValueError(
+                "Incorrect `chart_width` value. `chart_width` should be a float or int in the interval [10, inf]"
+            )
+        if self.chart_height < 10:
+            raise ValueError(
+                "Incorrect `chart_height` value. `chart_height` should be a float or int in the interval [10, inf]"
+            )
+
+    def _get_metric_names(self) -> tp.List[str]:
+        non_metric_columns = {Columns.Split, Columns.Model}
+        return [col for col in self.metrics_data.columns if col not in non_metric_columns]
+
+    def _get_folds_number(self) -> int:
+        return self.metrics_data[Columns.Split].nunique()
+
+    def _make_chart_data(self, fold_number: int) -> pd.DataFrame:
+        return self.metrics_data[self.metrics_data[Columns.Split] == fold_number].drop(columns=Columns.Split)
+
+    def _make_chart_data_avg(self) -> pd.DataFrame:
+        return self.metrics_data.drop(columns=Columns.Split).groupby(Columns.Model, sort=False).mean().reset_index()
+
+    def _create_chart(self, avg_folds: bool, fold_number: int, metric_x: str, metric_y: str) -> go.Figure:
+        chart_data = self._make_chart_data_avg() if avg_folds else self._make_chart_data(fold_number)
+        fig = px.scatter(
+            chart_data,
+            x=metric_x,
+            y=metric_y,
+            color=Columns.Model,
+            width=self.chart_width,
+            height=self.chart_height,
+            color_discrete_sequence=self.color_discrete_sequence,
+        )
+        fig.layout.update(showlegend=self.show_legend)
+        return fig
+
+    def display(self) -> None:
+        """TODO: add dosctring"""
+        metrics_list = self._get_metric_names()
+        n_splits = self._get_folds_number()
+
+        metric_x = widgets.Dropdown(description="Metric X:", value=metrics_list[0], options=metrics_list)
+        metric_y = widgets.Dropdown(description="Metric Y:", value=metrics_list[-1], options=metrics_list)
+        container_metrics = widgets.HBox(children=[metric_x, metric_y])
+
+        use_avg = widgets.Checkbox(description="Avg folds", value=True)
+        fold_number = widgets.Dropdown(description="Fold number:", value=0, options=list(range(n_splits)))
+        container_folds = widgets.HBox(children=[use_avg, fold_number])
+
+        scatter_chart = widgets.Output()
+
+        def update_chart(*args: tp.Any) -> None:
+            with scatter_chart:
+                scatter_chart.clear_output(wait=True)
+                fig = self._create_chart(use_avg.value, fold_number.value, metric_x.value, metric_y.value)
+                fig.show()
+
+        metric_x.observe(update_chart, "value")
+        metric_y.observe(update_chart, "value")
+        use_avg.observe(update_chart, "value")
+        fold_number.observe(update_chart, "value")
+
+        update_chart()
+        display(widgets.VBox([container_folds, container_metrics, scatter_chart]))
--- a/tests/visuals/test_metrics_app.py
+++ b/tests/visuals/test_metrics_app.py
@@ -1,0 +1,89 @@
+import typing as tp
+
+import pandas as pd
+import plotly.express as px
+import pytest
+
+from rectools import Columns
+from rectools.visuals.metrics_app import MetricsApp
+
+DF_METRICS = pd.DataFrame(
+    {
+        Columns.Model: ["Model1", "Model2", "Model1", "Model2", "Model1", "Model2"],
+        Columns.Split: [0, 0, 1, 1, 2, 2],
+        "prec@10": [0.1, 0.2, 0.3, 0.4, 0.5, 0.6],
+        "recall": [0.1, 0.2, 0.3, 0.4, 0.5, 0.6],
+    }
+)
+
+DF_METRICS_EMPTY = pd.DataFrame(
+    {
+        Columns.Model: ["Model1", "Model2", "Model1", "Model2", "Model1", "Model2"],
+        Columns.Split: [0, 0, 1, 1, 2, 2],
+    }
+)
+
+DF_METRICS_NO_MODEL = pd.DataFrame(
+    {
+        Columns.Split: [0, 0, 1, 1, 2, 2],
+        "prec@10": [0.1, 0.2, 0.3, 0.4, 0.5, 0.6],
+        "recall": [0.1, 0.2, 0.3, 0.4, 0.5, 0.6],
+    }
+)
+
+
+class TestMetricsApp:
+    @pytest.mark.parametrize("chart_width", (10, 100.0, 1000))
+    @pytest.mark.parametrize("chart_height", (10, 100.0, 1000))
+    @pytest.mark.parametrize("show_legend", (True, False))
+    @pytest.mark.parametrize(
+        "color_discrete_sequence",
+        (
+            None,
+            px.colors.qualitative.Plotly,
+            ["#0d0887", "#46039f", "#7201a8", "#9c179e", "#bd3786", "#d8576b"],
+        ),
+    )
+    def test_correct_run(
+        self,
+        chart_width: tp.Union[float, int],
+        chart_height: tp.Union[float, int],
+        show_legend: bool,
+        color_discrete_sequence: tp.Optional[tp.List[str]],
+    ) -> None:
+        MetricsApp.construct(
+            df_metrics_data=DF_METRICS,
+            chart_width=chart_width,
+            chart_height=chart_height,
+            show_legend=show_legend,
+            auto_display=False,
+            color_discrete_sequence=color_discrete_sequence,
+        )
+
+    def test_incorrect_width(self) -> None:
+        with pytest.raises(ValueError):
+            MetricsApp.construct(
+                df_metrics_data=DF_METRICS,
+                chart_width=0,
+                auto_display=False,
+            )
+
+    def test_incorrect_height(self) -> None:
+        with pytest.raises(ValueError):
+            MetricsApp.construct(
+                df_metrics_data=DF_METRICS,
+                chart_height=0,
+                auto_display=False,
+            )
+
+    def test_no_metric_columns(self) -> None:
+        with pytest.raises(KeyError):
+            MetricsApp.construct(
+                df_metrics_data=DF_METRICS_EMPTY,
+            )
+
+    def test_no_base_columns(self) -> None:
+        with pytest.raises(KeyError):
+            MetricsApp.construct(
+                df_metrics_data=DF_METRICS_NO_MODEL,
+            )
--- a/poetry.lock
+++ b/poetry.lock
@@ -1889,6 +1889,22 @@ test = ["appdirs (==1.4.4)", "covdefaults (>=2.3)", "pytest (>=7.4.3)", "pytest-
 type = ["mypy (>=1.8)"]
 
 [[package]]
+name = "plotly"
+version = "5.22.0"
+description = "An open-source, interactive data visualization library for Python"
+category = "main"
+optional = false
+python-versions = ">=3.8"
+files = [
+    {file = "plotly-5.22.0-py3-none-any.whl", hash = "sha256:68fc1901f098daeb233cc3dd44ec9dc31fb3ca4f4e53189344199c43496ed006"},
+    {file = "plotly-5.22.0.tar.gz", hash = "sha256:859fdadbd86b5770ae2466e542b761b247d1c6b49daed765b95bb8c7063e7469"},
+]
+
+[package.dependencies]
+packaging = "*"
+tenacity = ">=6.2.0"
+
+[[package]]
 name = "pluggy"
 version = "1.5.0"
 description = "plugin and hook calling mechanisms for python"
@@ -2572,6 +2588,22 @@ files = [
 ]
 
 [[package]]
+name = "tenacity"
+version = "8.3.0"
+description = "Retry code until it succeeds"
+category = "main"
+optional = false
+python-versions = ">=3.8"
+files = [
+    {file = "tenacity-8.3.0-py3-none-any.whl", hash = "sha256:3649f6443dbc0d9b01b9d8020a9c4ec7a1ff5f6f3c6c8a036ef371f573fe9185"},
+    {file = "tenacity-8.3.0.tar.gz", hash = "sha256:953d4e6ad24357bceffbc9707bc74349aca9d245f68eb65419cf0c249a1949a2"},
+]
+
+[package.extras]
+doc = ["reno", "sphinx"]
+test = ["pytest", "tornado (>=4.5)", "typeguard"]
+
+[[package]]
 name = "threadpoolctl"
 version = "3.4.0"
 description = "threadpoolctl"
@@ -2969,4 +3001,4 @@ visuals = ["ipywidgets"]
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.8.1, <3.13"
-content-hash = "583ec06080e8dec8c1fce878fbbce86e6ba578f343f5af55f8e76cadf3f35379"
+content-hash = "ff93666d9931b21a06288c2a80bb6b4703330b66628a299345214c45bcd53a5d"
--- a/rectools/columns.py
+++ b/rectools/columns.py
@@ -26,6 +26,7 @@ class Columns:
     Rank = "rank"
     Score = "score"
     Model = "model"
+    Split = "i_split"
     UserItem = [User, Item]
     Interactions = [User, Item, Weight, Datetime]
     Recommendations = [User, Item, Score, Rank]


tests/visuals/test_metrics_app.py
```diff
@@ -0,0 +1,89 @@
...
+    @pytest.mark.parametrize("show_legend", (True, False))
+    @pytest.mark.parametrize(
+        "color_discrete_sequence",
+        (
+            None,
+            px.colors.qualitative.Plotly,
+            ["#0d0887", "#46039f", "#7201a8", "#9c179e", "#bd3786", "#d8576b"],
+        ),
+    )
+    def test_correct_run(
```
blondered, line 47: We call it "test_happy_path"

blondered (changes requested):

pyproject.toml
```diff
@@ -77,18 +77,19 @@ torch = {version = ">=1.6.0, <3.0.0", optional = true}
 pytorch-lightning = {version = ">=1.6.0, <3.0.0", optional = true}
 
 ipywidgets = {version = ">=7.7,<8.2", optional = true}
+plotly = "^5.22.0"
```
feldlime, line 80: Please mark it optional since it's in extras (and run `poetry lock` after it)

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,128 @@
...
+    Jupyter widgets app for metric visualization and comparison.
+    Do not create instances of this class directly. Use `MetricsApp.construct` method instead.
+    """
+
+    def __init__(
+        self,
+        df_metrics_data: pd.DataFrame,
+        chart_width: tp.Union[float, int] = 800,
+        chart_height: tp.Union[float, int] = 600,
+        show_legend: bool = False,
```
feldlime, line 23: let's change to `True` here and in `construct`

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,128 @@
...
+        if self.chart_height < 10:
+            raise ValueError(
+                "Incorrect `chart_height` value. `chart_height` should be a float or int in the interval [10, inf]"
+            )
+
+    def _get_metric_names(self) -> tp.List[str]:
+        non_metric_columns = {Columns.Split, Columns.Model}
+        return [col for col in self.metrics_data.columns if col not in non_metric_columns]
+
+    def _get_folds_number(self) -> int:
```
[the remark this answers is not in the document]
feldlime: Don't agree, property assumes very fast calculation, here it may potentially be slow

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,128 @@
...
+        return self.metrics_data[Columns.Split].nunique()
+
+    def _make_chart_data(self, fold_number: int) -> pd.DataFrame:
+        return self.metrics_data[self.metrics_data[Columns.Split] == fold_number].drop(columns=Columns.Split)
+
+    def _make_chart_data_avg(self) -> pd.DataFrame:
+        return self.metrics_data.drop(columns=Columns.Split).groupby(Columns.Model, sort=False).mean().reset_index()
+
+    def _create_chart(self, avg_folds: bool, fold_number: int, metric_x: str, metric_y: str) -> go.Figure:
```
feldlime, line 87: 1. You don't need `fold_number` if `avg_folds is True`, so let it be optional
2. It's better to order arguments by importance, so here it should be `metric_x, metric_y, avg_folds, fold_number`

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,128 @@
...
+class MetricsApp:
+    """
+    Jupyter widgets app for metric visualization and comparison.
+    Do not create instances of this class directly. Use `MetricsApp.construct` method instead.
+    """
+
+    def __init__(
+        self,
+        df_metrics_data: pd.DataFrame,
+        chart_width: tp.Union[float, int] = 800,
```
feldlime, line 21: 1. How it can be float?
2. Plotly allows you to pass `None`, so it's better to use it here as well
3. In order not to list here all the plotly params and in the same time give the user maximum flexibility we could simply make a param `plotly_kwargs: Dict[str, Any]`

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,128 @@
...
+        """TODO: add dosctring"""
+        metrics_list = self._get_metric_names()
+        n_splits = self._get_folds_number()
+
+        metric_x = widgets.Dropdown(description="Metric X:", value=metrics_list[0], options=metrics_list)
+        metric_y = widgets.Dropdown(description="Metric Y:", value=metrics_list[-1], options=metrics_list)
+        container_metrics = widgets.HBox(children=[metric_x, metric_y])
+
+        use_avg = widgets.Checkbox(description="Avg folds", value=True)
+        fold_number = widgets.Dropdown(description="Fold number:", value=0, options=list(range(n_splits)))
```
feldlime, line 111: It's better to hide this field somehow if `Avg folds` is selected.

Or we can make a single Dropdown `Fold number` with options `Avg, 0, 1, 2, ...`

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## Unreleased
+
+### Added
+- Plotly scatterplot widget for metric-to-metric comparison ([#140](https://github.com/MobileTeleSystems/RecTools/pull/140))
```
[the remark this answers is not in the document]
Gooogr: Done

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,128 @@
...
+        if self.chart_width < 10:
+            raise ValueError(
+                "Incorrect `chart_width` value. `chart_width` should be a float or int in the interval [10, inf]"
+            )
+        if self.chart_height < 10:
+            raise ValueError(
+                "Incorrect `chart_height` value. `chart_height` should be a float or int in the interval [10, inf]"
+            )
+
+    def _get_metric_names(self) -> tp.List[str]:
```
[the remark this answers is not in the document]
Gooogr: Done

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,128 @@
...
+            raise ValueError("Incorrect input type. `metrics_data` should be a DataFrame")
+        if Columns.Split not in self.metrics_data.columns:
+            raise KeyError("Missing 'Split' column in `metrics_data` DataFrame")
+        if Columns.Model not in self.metrics_data.columns:
+            raise KeyError("Missing 'Model' column in `metrics_data` DataFrame")
+        if len(self.metrics_data.columns) < 3:
+            raise KeyError("`metrics_data` DataFrame assumed to have at least one metric column")
+
+    def _validate_input_patameters(self) -> None:
+        if self.chart_width < 10:
```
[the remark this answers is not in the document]
Gooogr: Outdated due to refactoring

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,128 @@
...
+        use_avg = widgets.Checkbox(description="Avg folds", value=True)
+        fold_number = widgets.Dropdown(description="Fold number:", value=0, options=list(range(n_splits)))
+        container_folds = widgets.HBox(children=[use_avg, fold_number])
+
+        scatter_chart = widgets.Output()
+
+        def update_chart(*args: tp.Any) -> None:
+            with scatter_chart:
+                scatter_chart.clear_output(wait=True)
+                fig = self._create_chart(use_avg.value, fold_number.value, metric_x.value, metric_y.value)
```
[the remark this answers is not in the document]
Gooogr: Done

tests/visuals/test_metrics_app.py
Gooogr: Done

pyproject.toml
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done, but may require additional discussion due to refactoring

rectools/visuals/metrics_app.py
Gooogr: I found that in Plotly's source code. However, the suggestion to switch to kwargs was great, so I've implemented it.

rectools/visuals/metrics_app.py
Gooogr: Done

Gooogr:
- Update CHANGELOG.md
- Update poetry files
- Update code according to the comments above

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,120 @@
...
+        metric_y = widgets.Dropdown(description="Metric Y:", options=self.metric_names, value=self.metric_names[-1])
+        use_avg = widgets.Checkbox(description="Avg folds", value=True)
+        fold = widgets.Dropdown(description="Fold number:", options=list(range(self.n_folds)), value=0)
+        scatter_chart = widgets.Output()
+
+        metric_x.observe(lambda change: self._update_chart(metric_x, metric_y, use_avg, fold, scatter_chart), "value")
+        metric_y.observe(lambda change: self._update_chart(metric_x, metric_y, use_avg, fold, scatter_chart), "value")
+        use_avg.observe(lambda change: self._update_chart(metric_x, metric_y, use_avg, fold, scatter_chart), "value")
+        use_avg.observe(lambda change: self._toggle_fold_number_visibility(fold, use_avg), "value")
+        fold.observe(lambda change: self._update_chart(metric_x, metric_y, use_avg, fold, scatter_chart), "value")
```
feldlime, line 110: since you're using the same function 4 times it makes sense to separate it

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _toggle_fold_number_visibility(self, fold: widgets.Dropdown, use_avg: widgets.Checkbox) -> None:
+        fold.layout.visibility = "hidden" if use_avg.value else "visible"
+
+    def display(self) -> None:
+        """TODO: add dosctring"""
+        metric_x = widgets.Dropdown(description="Metric X:", options=self.metric_names, value=self.metric_names[0])
+        metric_y = widgets.Dropdown(description="Metric Y:", options=self.metric_names, value=self.metric_names[-1])
+        use_avg = widgets.Checkbox(description="Avg folds", value=True)
+        fold = widgets.Dropdown(description="Fold number:", options=list(range(self.n_folds)), value=0)
```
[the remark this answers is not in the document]
blondered: Default value (as `value=` here) should also be taken from the `Columns.Split` first value from sorted list because users can provide metrics data calculated outside RecTools

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,120 @@
...
+
+class MetricsApp:
+    """
+    Jupyter widgets app for metric visualization and comparison.
+    Do not create instances of this class directly. Use `MetricsApp.construct` method instead.
+    """
+
+    def __init__(
+        self,
+        df_metrics_data: pd.DataFrame,
```
blondered, line 23: Let's call it either `metrics_data` or `metrics_df` or (even better I think) `models_metrics`. Here and in `construct`

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,120 @@
...
+        auto_display: bool = True,
+        plotly_kwargs: tp.Optional[tp.Dict[str, tp.Any]] = None,
+    ):
+        self.metrics_data = df_metrics_data
+        self.show_legend = show_legend
+        self.auto_display = auto_display
+        self.default_plotly_kwargs = {"width": DEFAULT_WIDTH, "height": DEFAULT_HEIGHT}
+        self.plotly_kwargs = {**self.default_plotly_kwargs, **(plotly_kwargs if plotly_kwargs is not None else {})}
+
+        self.fig = go.Figure()
```
blondered, line 34: Do we need this only for linters or is it actually making something? I'm just curious, didn't work with plotly that much

rectools/visuals/metrics_app.py
feldlime: IMO it's kinda obvious that metrics are the model metrics.  
I agree that `data` is redundant here.  
And IMO `df` is also redundant since we have the type annotation.  
Let's keep just `metrics`

feldlime: hmm, ok, makes a bit sense  
Let's do `model_metrics`

rectools/visuals/metrics_app.py
Gooogr: Only for mypy, it can't handle default None initialization

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,120 @@
...
+        self,
+        df_metrics_data: pd.DataFrame,
+        show_legend: bool = True,
+        auto_display: bool = True,
+        plotly_kwargs: tp.Optional[tp.Dict[str, tp.Any]] = None,
+    ):
+        self.metrics_data = df_metrics_data
+        self.show_legend = show_legend
+        self.auto_display = auto_display
+        self.default_plotly_kwargs = {"width": DEFAULT_WIDTH, "height": DEFAULT_HEIGHT}
```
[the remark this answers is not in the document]
Gooogr: I added this default option because of chart trembling after first interaction with any widget.
I agree that this approach seems dirty, so I've update code, now this width and height kwargs directly used only in _update_chart method.

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
Gooogr: Sure, done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done

Gooogr:
Updated the code based on feedback and added  docstrings. Remaining tasks:
* Create a Jupyter notebook example.
* Update the poetry lock file.

rectools/visuals/metrics_app.py
feldlime: I meant another thing here: I see 4 times repeated `lambda change: self._update_chart(metric_x, metric_y, use_avg, fold, scatter_chart)`.  And it make sense to convert it into a separate named function

feldlime: I meant other thing here: I see 4 times repeated `lambda change: self._update_chart(metric_x, metric_y, use_avg, fold, scatter_chart)`.  And it make sense to convert it into a separate named function

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## Unreleased
+
+### Added
+- `MetricsApp` to create plotly scatterplot widgets for metric-to-metric trade-off analysis ([#140](https://github.com/MobileTeleSystems/RecTools/pull/140))
```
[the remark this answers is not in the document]
feldlime: It's time to do this

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,158 @@
...
+    ) -> None:
+        with scatter_chart:
+            scatter_chart.clear_output(wait=True)
+            chart_data = self._make_chart_data_avg() if use_avg.value else self._make_chart_data(fold.value)
+            self.fig = px.scatter(
+                data_frame=chart_data,
+                x=metric_x.value,
+                y=metric_y.value,
+                color=Columns.Model,
+                width=self.plotly_kwargs.get("width", DEFAULT_WIDTH),
```
feldlime, line 121: This will cause an error like `multiple values for argument 'width'` if `width` in the `self.plotly_kwargs`

pyproject.toml
```diff
@@ -77,18 +77,19 @@ torch = {version = ">=1.6.0, <3.0.0", optional = true}
...
 
 ipywidgets = {version = ">=7.7,<8.2", optional = true}
+plotly = {version="^5.22.0", optional = true}
 
 
 [tool.poetry.extras]
 lightfm = ["lightfm"]
 nmslib = ["nmslib"]
 torch = ["torch", "pytorch-lightning"]
-visuals = ["ipywidgets"]
```
blondered, line 86: I've just realised that we need to add `MetricsApp` to compat.py https://github.com/MobileTeleSystems/RecTools/blob/main/rectools/compat.py and test https://github.com/MobileTeleSystems/RecTools/blob/main/tests/test_compat.py

blondered (changes requested):

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,158 @@
...
+                - `Columns.Split` - fold number
+                - Any other numeric columns which represent metric values
+        show_legend : bool, default True
+            Specifies whether to display the chart legend.
+        auto_display : bool, default True
+            Automatically displays the widgets immediately after initialization.
+        plotly_kwargs : tp.Optional[tp.Dict[str, tp.Any]], optional, default None
+            Additional arguments for customizing the `px.scatter` plot.
+
+        Returns
```
blondered, line 63: It also would be great to add `Examples` here with one simple example. This will be rendered in documentation just like here: https://rectools.readthedocs.io/en/stable/api/rectools.visuals.visual_app.VisualApp.html

blondered (changes requested):

CHANGELOG.md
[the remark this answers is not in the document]
Gooogr: Done

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
Gooogr: Probably outdated, because I removed `self.default_plotly_kwargs` from current version

rectools/visuals/metrics_app.py
Gooogr: Probably outdated

rectools/visuals/metrics_app.py
Gooogr: Sorry, I don't get your idea. How dictionary could possibly have multiple values for one key? Anyway, this part is also outdated now.

Gooogr:
Changed the principle of widget design, now it relies on `go.WidgetFigure`. Also switch from plotly express to plotly graph object scatter chart, because WidgetFigure doesn't support express wrapper.
TODO:
* Expand `MetricsApp` documentation
* Add `MetricsApp` to `compat.py`
* Probably add width and height as optional parameters again.

Currently, support for `plotly_kwargs` has been removed due to the large amount of customization inside the `display` method. But the users would be able to customize the chart appearance at the static rendering stage. Add example in jupyter notebook.

rectools/visuals/metrics_app.py
Gooogr: Fixed, thank you.

Gooogr:
Update MetricsApp

* Refactor display method
* Add jupyter example
* Add docstring example
* Update tests

Add optional parameters support of plotly.go.Layout

rectools/visuals/__init__.py
```diff
@@ -22,11 +22,13 @@
 ---------
 `visuals.VisualApp` - Jupyter app for visual comparison of recommendations
 `visuals.ItemToItemVisualApp` - Jupyter app for visual comparison of item-to-item recommendations
+`visuals.MetricsApp` - Jupyter app for visual metrics comparison
 """
 
 try:
+    from .metrics_app import MetricsApp
```
blondered, line 29: It is possible that user installed ipywidgets but didn't install plotly. In this case we need to give him full version of VisualApp, not dummy one from `compat`. And a dummy version of MetricsApp.
So we need to have separate try/except statements for imports of these classes

pyproject.toml
```diff
@@ -125,4 +126,4 @@ target-version = ["py38", "py39", "py310", "py311", "py312"]
 
 [build-system]
 requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
+build-backend = "poetry.core.masonry.api"
```
blondered, line 129: No new line at end of file

rectools/visuals/metrics_app.py
feldlime: You probably can do smth like [this](https://github.com/MobileTeleSystems/RecTools/blob/main/rectools/models/implicit_als.py#L64) - specify the type and don't specify the value

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,212 @@
...
+        >>> fig = app.fig
+        >>> fig = fig.update_layout(
+        ...    title="Metrics: recall@10 vs novelty@10",
+        ...    xaxis_title="novelty@10",
+        ...    yaxis_title="recall@10",
+        ...    legend_title="Models",
+        ... )
+        """
+        app = cls(models_metrics, show_legend, auto_display, layout_kwargs)
+        if auto_display:
```
feldlime, line 96: shouldn't be it in `__init__`?

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,212 @@
...
+        return app
+
+    @property
+    def metric_names(self) -> tp.List[str]:
+        """List of metric column names from the `models_metrics`."""
+        non_metric_columns = {Columns.Split, Columns.Model}
+        return [col for col in self.models_metrics.columns if col not in non_metric_columns]
+
+    @property
+    def model_names(self) -> tp.List[str]:
```
feldlime, line 107: You're using this property quite often, but they theoretically may be slow and may cause problems during the chart updates. Let's add `lru_cache` here.
(Maybe for prev. and next one as well)

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,212 @@
...
+        if not isinstance(self.models_metrics, pd.DataFrame):
+            raise ValueError("Incorrect input type. `metrics_data` should be a DataFrame")
+        if Columns.Split not in self.models_metrics.columns:
+            raise KeyError("Missing `Split` column in `metrics_data` DataFrame")
+        if Columns.Model not in self.models_metrics.columns:
+            raise KeyError("Missing `Model`` column in `metrics_data` DataFrame")
+        if len(self.models_metrics.columns) < 3:
+            raise KeyError("`metrics_data` DataFrame assumed to have at least one metric column")
+
+    def _make_chart_data(self, fold_number: int) -> pd.DataFrame:
```
feldlime, line 131: maybe it also makes sense to add caching for this and the next method

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,212 @@
...
+                    mode="markers",
+                    name=model,
+                    hovertemplate=template,
+                    showlegend=self.show_legend,
+                )
+            fig_widget.layout.xaxis.title = metric_x.value
+            fig_widget.layout.yaxis.title = metric_y.value
+        self.fig = go.Figure(fig_widget.data)
+
+    def _update_chart(
```
feldlime, line 163: this method looks very very similar to the `_create_chart`. It'd be nice to avoid the code duplication

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,212 @@
...
+            "legend_title": LEGEND_TITLE,
+        }
+        layout_params.update(self.layout_kwargs)
+        fig_widget = go.FigureWidget(layout=layout_params)
+
+        metric_x.observe(lambda upd: self._update_chart(fig_widget, metric_x, metric_y, use_avg, fold_i), "value")
+        metric_y.observe(lambda upd: self._update_chart(fig_widget, metric_x, metric_y, use_avg, fold_i), "value")
+        use_avg.observe(lambda upd: self._update_chart(fig_widget, metric_x, metric_y, use_avg, fold_i), "value")
+        use_avg.observe(lambda upd: self._update_fold_visibility(use_avg, fold_i), "value")
+        fold_i.observe(lambda upd: self._update_chart(fig_widget, metric_x, metric_y, use_avg, fold_i), "value")
```
feldlime, line 209: My prev. comment about duplicating lambdas still makes sense. Let's please move it to a separate function

pyproject.toml
Gooogr: Yes, but the same style pattern used in the main branch or any other branches: https://github.com/MobileTeleSystems/RecTools/blob/main/pyproject.toml#L128

So, is it really necessary in this case?

rectools/visuals/metrics_app.py
Gooogr: Test it and found out that `self.fig = go.Figure()` is better option. It works as fallback if user create app with option `auto_display = False` and try to get figure `app.fig` without any widget interactions. If we don't specify `self.fig` value this would lead to error. I updated `happy_path`tests to make sure that we can handle `fig` extraction.

rectools/visuals/__init__.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Make sense, done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done. Initially I try to keep create and update steps separately from each other, but yes, after all chart tuning functions used to look quite similar.

rectools/visuals/metrics_app.py
Gooogr: Unfortunately I can't find clean solution for this idea. We need to observe every widget and need to pass update methods through lambda function. 
We could significantly improve this part if we would keep all widgets as class variables, which I try to avoid to reduce coupling between display and data preparation logic.

pyproject.toml
blondered: I see you point. It' kinda strange it shows on the diff in this particular PR

blondered:
Let's add explicit info for new users about the use of MetricsApp in 2_cross_validation.ipynb:

1. "And let's visualuze metrics" ->
In RecTools we have interactive `MetricsApp` for detailed analysis of metrics trade-off between different models. 
`visuals` extension is required to run the path of code below. You can install it with `pip install rectools[visuals]`

2. `models_metrics.rename(columns={"model": Columns.Model, "i_split": Columns.Split})` - this we can drop, the names are already the same
3. `app = MetricsApp.construct(pd.DataFrame(cv_results["metrics"]))` for one line of code instead of two
4. Before app screenshot: "If you run this notebook, you will get **interactive** widgets with active buttons to select metrics and folds. For offline presentation we keep **static** screenshots of the actual app."
5. After app screenshot: "If you want to save static image from your app, you can access `plotly.Figure` and render it as an image. Make sure you have 'kaleido' package which will process rendering. You can install it with `pip install kaleido==0.2.1` and restart the kernel.

rectools/visuals/metrics_app.py
feldlime: You can write a small function just before metric_x.observe like 
```
def update(event: ...) -> None:
    self.update_...
```
and use it instead of lambda

rectools/visuals/metrics_app.py
```diff
@@ -0,0 +1,194 @@
...
+            Additional arguments from `plotly.graph_objects.Layout`
+
+        Returns
+        -------
+        MetricsApp
+            An instance of `MetricsApp`, providing interactive Jupyter widget for metric visualization.
+
+        Examples
+        --------
+        Create interactive widget
```
blondered, line 72: Examples in the documentation are not rendered correctly right now. You can check rendering for your PR here: https://rectools--140.org.readthedocs.build/en/140/api/rectools.visuals.metrics_app.MetricsApp.html
Me guess is that you are missing empty line between text (`Create interactive widget`) and code example. Let's try and add it.

When all is good examples will be rendered just like here: https://rectools--140.org.readthedocs.build/en/140/api/rectools.visuals.visual_app.VisualApp.html

Gooogr: Oh, sorry for that. I would add extensions to my VS code to check docs renders of future commits. Fixed

rectools/visuals/metrics_app.py
Gooogr: Done

blondered (approved):

PR #81, opened 2024-01-15
ARP metric added

mikesokolovv:
## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Optimization

PR #148, opened 2024-05-21
Added intersection metric (#95)

### LINKED ISSUE #95: Recommendations intersection metric
### Feature Description

Metric to measure intersection in user-item (or item-item) pairs between recommendation lists.

### Why this feature?

It helps for both candidate-generators selection in pipelines, and for popularity bias measurement.

### Additional context

```

intersecton = Intersection(k=10)

# one metric
intersecton_value = intersecton.calc(reco=recos, ref_recos=ref_recos)
intersecton_per_user = precision.calc_per_user(reco=recos, ref_recos=ref_recos)

# calc_metrics
calc_metrics(
    metrics = {"precision": precision, "intersection": intersecton},
    reco=recos,
    interactions=df_test,
    ref_recos: Union[pd.Datafame, Dict[Hashable, pd.DataFrame]]=ref_recos
)

```
We can keep ref_recos a simple pd.DataFrame to calculate intersections with one algorithm.
For multiple intersection calculations we can pass multiple models recommendations in a dict:
`ref_recos = {"one": ref_recos_one, "two": ref_recos_two}`
Result dict from `calc_metrics` will have `intersection_one` and `intersection_two` keys if ref_recos is a dict (merging keys from `metrics` and `ref_recos`.

```
# cross_validate
cv_results = cross_validate(
    dataset=dataset,
    splitter=splitter,
    models=models,
    metrics=metrics,
    k=10,
    filter_viewed=True,
    ref_models=["one", "two"], # here we just select keys from `models` argument
    validate_ref_models=False  # optionally exclude ref_models from other metrics calculation
)
```

azatnv:
## Description

- Added intersection metric. 

Closes #95

## Type of change

- [ ] Optimization

## How Has This Been Tested?


rectools/metrics/intersection.py
```diff
@@ -0,0 +1,71 @@
...
+            Values of metric (index - user id, values - metric value for every user).
+        """
+        self._check(reco)
+        self._check(ref_reco)
+
+        if not ref_k:
+            ref_k = ref_reco[Columns.Rank].max()
+        filtered_ref_reco = ref_reco[ref_reco[Columns.Rank] <= ref_k]
+
+        recall = Recall(k=self.k)
```
blondered, line 70: Nice catch to use Recall. But it't got to be the other way around. We need the share of `reco` that is present in `ref_reco`. So `reco` should be passed as `interactions` to `Recall`.
Also because of this we need to filter `reco` before metric calculation, not `ref_reco`
Also let's extensively explain this logic for users in the docstring

rectools/metrics/intersection.py
```diff
@@ -0,0 +1,71 @@
...
+
+        Returns
+        -------
+        pd.Series:
+            Values of metric (index - user id, values - metric value for every user).
+        """
+        self._check(reco)
+        self._check(ref_reco)
+
+        if not ref_k:
```
blondered, line 66: Discussed to change logic and equal it to` k` when not provided. Sorry for the delay

rectools/metrics/intersection.py
```diff
@@ -0,0 +1,71 @@
...
+
+from rectools import Columns
+from rectools.metrics.base import MetricAtK
+from rectools.metrics.classification import Recall
+
+
+@attr.s
+class Intersection(MetricAtK):
+    """
+    Metric to measure intersection in user-item (or item-item) pairs between recommendation lists.
```
blondered, line 14: There is no support for item-item case in the code. So let's delete this from the docstring before it is implemented.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Extended Theory&Practice RecSys baselines tutorial ([#139](https://github.com/MobileTeleSystems/RecTools/pull/139))
+- Intersection metric
```
blondered, line 13: Please follow the logic with link to the PR

rectools/metrics/intersection.py
```diff
@@ -0,0 +1,71 @@
+from typing import Optional
```
blondered, line 1: For implementing a new metric apart from the metric itself it is also necessary to implement some other pieces of code to make it work in the pipelines. This is a reference PR for you: https://github.com/MobileTeleSystems/RecTools/pull/81/files
What is missing now:
- `calc_intersection_metrics` in `intersection.py`
- `calc_intersection_metrics` call in `scoring.py`
- new metric added to `metrics.__init__.py`
- new metric added to `test_scoring.py`

tests/metrics/test_intersection.py
```diff
@@ -0,0 +1,48 @@
...
+
+    @pytest.mark.parametrize(
+        "k,ref_k,expected_users,expected_intersection",
+        (
+            (2, 2, [1, 2, 4, 5], [0.0, 1.0, 1 / 2, 1 / 3]),
+            (3, None, [1, 2, 3, 4, 5], [1 / 2, 1.0, 1 / 2, 2 / 3, 2 / 4]),
+            (3, 6, [1, 2, 3, 4, 5], [1 / 2, 1.0, 0.0, 1 / 2, 2 / 3]),
+        ),
+    )
+    def test_calc(self, k: int, ref_k: int, expected_users: List[int], expected_intersection: List[float]) -> None:
```
blondered, line 21: We always try to add test on corner cases for metrics. Here I think it is necessary to check that empty `ref_reco` will result in value 0 for the metric.

blondered (changes requested):

azatnv:
I have addressed all the comments

rectools/metrics/scoring.py
```diff
@@ -24,18 +24,20 @@
...
 from .diversity import DiversityMetric, calc_diversity_metrics
+from .intersection import IntersectionMetric, calc_intersection_metrics
 from .novelty import NoveltyMetric, calc_novelty_metrics
 from .popularity import PopularityMetric, calc_popularity_metrics
 from .ranking import RankingMetric, calc_ranking_metrics
 from .serendipity import SerendipityMetric, calc_serendipity_metrics
 
 
-def calc_metrics(  # noqa  # pylint: disable=too-many-branches
+def calc_metrics(  # noqa  # pylint: disable=too-many-branches,too-many-locals,too-many-statements
```
blondered, line 34: Thank you for added support in `calc_metrics` (and for the PR itself, great job!). One more thing before we can close this issue completely. Could you please add support of new metric to `cross_validate`?
Details can be found in the issue description.

blondered (changes requested):

rectools/metrics/intersection.py
```diff
@@ -0,0 +1,130 @@
...
+    k : int
+        Number of items in top of recommendations list that will be used to calculate metric.
+    ref_k : int, optional
+        Number of items in top of reference recommendations list that will be used to calculate metric.
+        If ``ref_k`` is None than ``ref_reco`` will be filtered with ``ref_k = k``. Default: None.
+    """
+
+    ref_k: Optional[int] = attr.ib(default=None)
+
+    def calc(self, reco: pd.DataFrame, ref_reco: pd.DataFrame) -> float:
```
blondered, line 39: It needs to be either `ref_reco` or `ref_recos` absolutely everywhere. Each variant is ok. We can keep `ref_reco` just to make it closer to `reco` argument

azatnv: Changed

rectools/model_selection/cross_validate.py
```diff
@@ -129,6 +136,11 @@ def cross_validate(  # pylint: disable=too-many-locals
         interactions_df_test = interactions.df.iloc[test_ids]  # 1x internal
         test_users = interactions_df_test[Columns.User].unique()  # 1x internal
         catalog = interactions_df_train[Columns.Item].unique()  # 1x internal
+        if ref_models is not None and validate_ref_models:
```
blondered, line 139: There is some misunderstanding of logic I think.

If ref_models is not None -> `ref_reco` dict needs to be collected from all ref models recommendations. Other models will receive intersection metric for their recommendations with these ref models recommendation.
If ref_models is None -> `ref_reco = None`

If `validate_ref_models` is set to False then ref models recommendations are only used for other models metrics. We do not compute metrics for these models themselves.
If `validate_ref_models` is set to True then ref models also get all metrics calculated apart from providing recommendations for intersections calculation.

I'm sorry if I didn't write it explicitly in the issue description. If you have any questions please feel free to ask them here or in telegram. I'll be happy to help.

rectools/model_selection/cross_validate.py
```diff
@@ -61,6 +61,8 @@ def cross_validate(  # pylint: disable=too-many-locals
     filter_viewed: bool,
     items_to_recommend: tp.Optional[ExternalIds] = None,
     prefer_warm_inference_over_cold: bool = True,
+    ref_models: tp.Optional[tp.List[str]] = None,
```
blondered, line 64: We'll need a "happy path" test of these parameters in `tests.model_selection.test_cross_validate`
Something like `test_happy_path_with_intersection`.

rectools/model_selection/cross_validate.py
```diff
@@ -90,6 +92,11 @@ def cross_validate(  # pylint: disable=too-many-locals
         Set to `True` to enable "warm" recommendations for all applicable models.
         Set to `False` to treat all new users and items as "cold" and not to provide features for them.
         If new users and items are filtered from test in splitter, this argument has no effect.
+    ref_models : list(str), optional
+        The keys from `models` argument to compute intersection metrics.
```
blondered, line 96:
```suggestion
        The keys from `models` argument to compute intersection metrics. These models recommendations will be used as `ref_reco` for other models intersection metrics calculation.
```

rectools/model_selection/cross_validate.py
```diff
@@ -90,6 +92,11 @@ def cross_validate(  # pylint: disable=too-many-locals
...
+        Obligatory only if `IntersectionMetric` instances present in `metrics`.
+    validate_ref_models : bool
+        If True include `ref_models` to other metrics calculation. Default: False.
```
blondered, line 99:
```suggestion
        If True include models specified in `ref_models` to all metrics calculations and receive their metrics from cross-validation. Default: False.
```

blondered (changes requested):

rectools/metrics/scoring.py
```diff
@@ -57,6 +59,11 @@ def calc_metrics(  # noqa  # pylint: disable=too-many-branches
     catalog : collection, optional
         Collection of unique item ids that could be used for recommendations.
         Obligatory only if `ClassificationMetric` or `SerendipityMetric` instances present in `metrics`.
+    ref_reco : Union[pd.DataFrame, Dict[Hashable, pd.DataFrame]], optional
+        Reference recommendations table(s) with columns `Columns.User`, `Columns.Item`, `Columns.Rank`.
+        For multiple intersection calculations we can pass multiple models recommendations in a dict:
+        ``ref_recos = {"one": ref_recos_one, "two": ref_recos_two}``
```
blondered, line 65:
```suggestion
        ``ref_reco = {"one": ref_reco_one, "two": ref_reco_two}``
```

blondered (changes requested):

tests/model_selection/test_cross_validate.py
```diff
@@ -268,3 +273,128 @@ def test_happy_path_with_features(self, prefer_warm_inference_over_cold: bool) -
...
+                    {"model": "random", "i_split": 1, "precision@2": 0.375, "recall@1": 0.25},
+                ],
+            ),
+            (
+                ["random", "popular"],
+                False,
+                [],
+            ),
+            (
+                ["random", "popular"],
```
blondered, line 346: Everything is right except for this case (and the problem is not in the test itself but in the logic in `cross_validate`).
When we have all models passed as ref_models and we have `validate_ref_models` set to True, we need to have intersections calculated for all models on all others. Instead of this here we get all intersections dropped for all models.

Nice test cases selection btw, very useful for this case.

rectools/model_selection/cross_validate.py
```diff
@@ -137,7 +204,35 @@ def cross_validate(  # pylint: disable=too-many-locals
...
+                internal_item_ids_to_recommend=item_ids_to_recommend,
+            )
+            models_ref_reco = {model_name: None if model_name in ref_models else ref_reco for model_name in models}
+
+            metrics_without_intersection = {
+                metric_name: metric
+                for metric_name, metric in metrics.items()
+                if not isinstance(metric, IntersectionMetric)
+            }
+            models_metrics = {
```
blondered, line 224: The problem in logic is here. Imagine the case when we have multiple `ref_models` and we have `validate_ref_models` set to True.
In this case we need to calculate all metrics (including intersection ones) for all models.
The only thing we can drop is calculating model intersection with itself (we can return `1` for this case)

blondered (changes requested):

rectools/model_selection/cross_validate.py
```diff
@@ -146,12 +241,14 @@ def cross_validate(  # pylint: disable=too-many-locals
                 filter_viewed=filter_viewed,
                 items_to_recommend=item_ids_to_recommend,
             )
+
             metric_values = calc_metrics(
-                metrics,
+                models_metrics[model_name],
```
blondered, line 246: I think we can keep all the metrics here for all models

rectools/model_selection/cross_validate.py
```diff
@@ -146,12 +241,14 @@ def cross_validate(  # pylint: disable=too-many-locals
...
                 reco=reco,
                 interactions=interactions_df_test,
                 prev_interactions=interactions_df_train,
                 catalog=catalog,
+                ref_reco=models_ref_reco[model_name],
```
blondered, line 251: Here we can pass `ref_reco` from `_get_ref_reco` func output. Only filter model name that is now validated from keys of the dict.

rectools/model_selection/cross_validate.py
```diff
@@ -137,7 +204,35 @@ def cross_validate(  # pylint: disable=too-many-locals
...
+            }
+        else:
+            models_ref_reco = {model_name: None for model_name in models}
+            models_metrics = {model_name: metrics for model_name in models}
+
         for model_name, model in models.items():
+            if ref_models is not None and model_name in ref_models and not validate_ref_models:
+                continue
+
             model.fit(fold_dataset)
```
blondered, line 236: here we can check if we have this model recos in `ref_reco` already. If we do -> get reco from there and not fit for the second time

rectools/metrics/intersection.py
```diff
@@ -0,0 +1,137 @@
...
+) -> Dict[str, float]:
+    """
+    Calculate intersection metrics.
+
+    Warning: It is not recommended to use this function directly.
+    Use `calc_metrics` instead.
+
+    Parameters
+    ----------
+    metrics : dict(str -> PopularityMetric)
```
blondered, line 112:
```suggestion
    metrics : dict(str -> IntersectionMetric)
```

rectools/metrics/intersection.py
```diff
@@ -0,0 +1,137 @@
...
+        recall = Recall(k=self.ref_k)
+
+        return recall.calc_per_user(ref_reco, filtered_reco[Columns.UserItem])
+
+
+IntersectionMetric = Intersection
+
+
+def calc_intersection_metrics(
+    metrics: Dict[str, Intersection],
```
blondered, line 100:
```suggestion
    metrics: Dict[str, IntersectionMetric],
```

blondered: And it's time to pull changed from `main`

blondered (approved):

Issue #98, opened 2024-02-21
DebiasWrapper for metrics


blondered:
### Feature Description

A metric wrapper that creates debiased validation in case of strong popularity bias **in test data**. One way to do this is to fight power-law popularity distribution in test interactions on each fold with down-sampling **fold** popular items.

### Why this feature?

It helps as a correct goal for hyper-parameters tuning and model selection

### Additional context

Algorithm to detect and down-sample excessively popular items. More algorithms and modifications can be proposed here. For now we can use IQR (interquartile-range) that is also used for boxplots: logic.
1. We find first and third quartiles in **test** items popularity distribution (Q1 and Q3)
2. IQR = Q3 - Q1. This is interquartile range. 50% of the observed data is inside this range.
3. Outliers popularity border will be defined as Q3 + iqr_coef * IQR
4. Maximum accepted popularity will be defined as the **maximum value inside** the border.
5. Every item that exceeds the border should be down-sampled to match the maximum accepted popularity. 

For all exceeding items in the test fold we need to randomly keep only the maximum allowed subset of users. We use downsampling for this.

The wrapper changes test interactions, but afterwards any metrics can be calculated as usual.

 ```
from rectools.metrics import DebiasWrapper, Precision

debiased_precision = DebiasWrapper(Precision(k=10), iqr_coef=1.5, random_state=32)
```

Other possible namings are: PopDownSamplingWrapper, DownSamplingWrapper, UnbiasedWrapper

PR #152, opened 2024-06-06
Feature/debias wrapper

In48semenov:
## Description

Added DebiasWrapper for metrics 
Closes [#98](https://github.com/orgs/MobileTeleSystems/projects/1/views/1?pane=issue&itemId=54005450)

## Type of change

- [ ] Optimization

## How Has This Been Tested?


rectools/metrics/debias.py
```diff
@@ -0,0 +1,118 @@
...
+
+    Returns
+    -------
+    pd.DataFrame
+        Downsampling interactions.
+    """
+    if len(interactions) == 0:
+        return interactions
+
+    item_popularity = interactions[Columns.Item].value_counts()
```
feldlime, line 84: What if there are multiple rows with the same user+item pairs?  Don't we want to use the same approach of popularity calculation we used in popularity models?

rectools/metrics/debias.py
```diff
@@ -0,0 +1,118 @@
...
+
+    quantiles = item_popularity.quantile(q=[0.25, 0.75])
+    q1, q3 = quantiles.loc[0.25], quantiles.loc[0.75]
+    iqr = q3 - q1
+    max_border = int(q3 + debias_config.iqr_coef * iqr)
+
+    item_outside_max_border = item_popularity[item_popularity > max_border].index
+
+    interactions_result = interactions[~interactions[Columns.Item].isin(item_outside_max_border)]
+    interactions_downsampling = interactions[interactions[Columns.Item].isin(item_outside_max_border)]
```
feldlime, line 94: Please avoid code duplication

rectools/metrics/debias.py
```diff
@@ -0,0 +1,118 @@
...
+
+    interactions_downsampling = (
```
feldlime, line 96: I think we can speed this up (please check my code is doing what you want it to do)
image

rectools/metrics/debias.py
```diff
@@ -0,0 +1,118 @@
...
+        interactions_downsampling.groupby(Columns.Item, as_index=False)[Columns.User]
+        .agg(lambda users: users.sample(max_border, random_state=debias_config.random_state).tolist())
+        .explode(Columns.User)
+    )
+
+    interactions_result = pd.concat([interactions_result, interactions_downsampling]).sort_values(
```
feldlime, line 102: Why do we need to sort them?

rectools/metrics/debias.py
```diff
@@ -0,0 +1,118 @@
...
+        Columns.User, ignore_index=True
+    )
+
+    interactions_result[Columns.User] = interactions_result[Columns.User].astype(interactions[Columns.User].dtypes)
```
feldlime, line 106: Why can we loose types?

rectools/metrics/debias.py
```diff
@@ -0,0 +1,118 @@
...
+    interactions_result[Columns.Item] = interactions_result[Columns.Item].astype(interactions[Columns.Item].dtypes)
+
+    if Columns.Rank in interactions.columns:
```
feldlime, line 109: Why can we have rank in interactions?

tests/metrics/test_debias.py
```diff
@@ -0,0 +1,243 @@
...
+            HitRate(k=3),
+            MAP(k=1),
+            MAP(k=3),
+            NDCG(k=1),
+            NDCG(k=3),
+            MRR(k=1),
+            MRR(k=3),
+        ),
+    )
+    def test_debias_metric_calc(
```
feldlime, line 140: this and next tests should be in `test_scoring.py`

rectools/metrics/classification.py
```diff
@@ -392,7 +419,10 @@ def _calc_per_user_from_confusion_df(self, confusion_df: pd.DataFrame) -> pd.Ser
 
 
 def calc_classification_metrics(
-    metrics: tp.Dict[str, tp.Union[ClassificationMetric, SimpleClassificationMetric]],
+    metrics: tp.Dict[
+        str,
+        tp.Union[ClassificationMetric, SimpleClassificationMetric],
+    ],
```
feldlime, line 425: Why this line is split into several lines?

rectools/metrics/classification.py
```diff
@@ -430,15 +460,25 @@ def calc_classification_metrics(
         If unexpected metric is present in `metrics`.
     """
     k_map = defaultdict(list)
+    merged_debias = {}
     for name, metric in metrics.items():
         k_map[metric.k].append(name)
+        if not isinstance(metric, (ClassificationMetric, SimpleClassificationMetric)):
```
feldlime, line 466: why do we need this check?

rectools/metrics/classification.py
```diff
@@ -430,15 +460,25 @@ def calc_classification_metrics(
...
+            raise TypeError(f"Unexpected classification metric {metric}")
+        if metric.debias_config is not None and metric.debias_config.keyname not in merged_debias:
+            merged_debias[metric.debias_config.keyname] = make_downsample(merged, metric.debias_config)
```
feldlime, line 469: You can make the config frozen (which actually makes sense) and use it directly as a key

rectools/metrics/debias.py
```diff
@@ -0,0 +1,118 @@
...
+    def keyname(self) -> str:
+        """
+        Compose a config's key name based on the iqr_coef and random_state parameters.
+
+        Returns
+        -------
+        str
+            String representing the key name in the format `"{iqr_coef}_{random_state}"`.
+        """
+        return f"{self.iqr_coef}_{self.random_state}"
```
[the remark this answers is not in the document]
feldlime: Ignore this comment if you agree to remove this property at all (check other comments)

rectools/metrics/ranking.py
```diff
@@ -37,8 +40,12 @@ class _RankingMetric(MetricAtK):
     ----------
     k : int
         Number of items at the top of recommendations list that will be used to calculate metric.
+    debias_config : DebiasConfig, default None
+        Config with debias method parameters (iqr_coef, random_state).
     """
 
+    debias_config: DebiasConfig = attr.ib(default=None)
```
feldlime, line 47: Probably we can add this to `MetricAtK`? Together with the code for handling it

rectools/metrics/classification.py
```diff
@@ -85,6 +90,9 @@ def calc_per_user(self, reco: pd.DataFrame, interactions: pd.DataFrame, catalog:
         pd.Series
             Values of metric (index - user id, values - metric value for every user).
         """
+        if self.debias_config is not None:
```
feldlime, line 93: I think it's really unclear and unsafe that  we execute downsampling in `calc_per_user`, but not in `calc_per_user_from_confusion_df`. 
User would probably like to use `calc_per_user_from_confusion_df` themselves but they wouldn't expect debias is not executed there

rectools/metrics/ranking.py
blondered: Debias is not applicable to all of the metrics. I suggest to add it only to appicable types of metrics.

rectools/metrics/debias.py
In48semenov: **Done**

rectools/metrics/debias.py
In48semenov: correct now?

rectools/metrics/debias.py
In48semenov: **Done**
But there is more randomness in applying the debias method, so there may be problems with tests.

rectools/metrics/ranking.py
feldlime: Let's  then probably add a class like `DibiasableMetrikAtK`?

feldlime: Let's  then probably add a class like `DibiasableMetrikAtK`?

rectools/metrics/debias.py
```diff
@@ -0,0 +1,118 @@
...
+        Coefficient for defining as the maximum value inside the border.
+    random_state : float, default 32
+        Pseudorandom number generator state to control the down-sampling.
+    """
+
+    iqr_coef: float = attr.ib(default=1.5)
+    random_state: int = attr.ib(default=32)
+
+    @property
+    def keyname(self) -> str:
```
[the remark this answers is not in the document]
In48semenov: **Done**
I removed this property and made `DebiasConfig` as frozen

tests/metrics/test_debias.py
feldlime: Sorry, my bad.  
The point was that in `test_debias.py` should be only tests that test the code from `debias.py`. And if you want to test `Precision` metric with debias, the test should be in `test_classification.py` since `Precision` depends on debias, not vice versa

rectools/metrics/debias.py
feldlime: It's a point to discuss.  I think now it's move correct.  But probably it'd be nice to have different options, and it also makes sense to keep the things in one style.  
@blondered wdyt?

rectools/metrics/classification.py
In48semenov: This is the [branch in the condition](https://github.com/MobileTeleSystems/RecTools/blob/main/rectools/metrics/classification.py#L449)

rectools/metrics/debias.py
In48semenov: It's the same as interactions, but with an extra `rank` column.
Снимок экрана 2024-06-07 в 19 12 01

I added this code, because the separate methods for ranking metrics and classification metrics functions work directly with `merge_reco`. In addition, the structure of the resulting `merge_reco` dataframe doesn't conflict with the debias calculation.

rectools/metrics/ranking.py
In48semenov: I don't think it will solve this point, as there will also be classification and ranking metrics inherited from it, which can be and are used without the debias mechanism

tests/metrics/test_debias.py
In48semenov: I agree. I'll rewrite the tests

rectools/metrics/debias.py
In48semenov: sorry, it's my mistake.
Because I forgot that I made the `random_state` parameter in `DebiasConfig` as `None`, so the results of the `make_debias` function changed in the tests.

rectools/metrics/debias.py
blondered: My main point: in all applicable for debias models we guarantee that multiple rows for the same user-item pair do not have effect on metric value. They are actually dropped (since we need them only to assign TP to reco pairs).
Because of this we can actually deduplicate before downsampling. And then take simple value_counts. It kinda guarantees correct downsampling.

rectools/metrics/ranking.py
blondered: @In48semenov what are the examples? I think we can probably debias anything that belongs to classification and ranking

In48semenov: Sorry, I misunderstood the previous comments

Yes, `DibiasableMetrikAtK` class would look quite reasonable, as it would be easy to add new debias methods to it in the future

rectools/metrics/debias.py
feldlime: So, now it's fine? :)

rectools/metrics/debias.py
feldlime: Ok, anyway, it seems that all the columns from the original `interactions` table have to be present in the `interactions_result` without any additional merges, no?

rectools/metrics/classification.py
feldlime: Yes, but it's not necessary, added just in case.  

However, double-checking here is really too much. Let's drop at least the last branch in the 2nd check

rectools/metrics/classification.py
feldlime: Sounds reasonable, but you will need to call this private method from `calc_classification_metrics`, right? Not the best practice

rectools/metrics/debias.py
In48semenov: Yes)

rectools/metrics/classification.py
blondered: From `calc_classification_metrics`, yes. What do you suggest?
Another option is to make `calc_per_user_from_confusion_df` public and write a docstring. I don't see other ways.

blondered: Or we can add boolean argument, smth like "is_confusion_df_debiased".
If it is None or False -> Check is debias_config is not None -> through error

rectools/metrics/debias.py
blondered: We've discussed that we need to skip 'popularity' word in all docstrings and code. It is kinda confusing. Instead let's call it "number of users that interacted with item". Let's please do it everywhere and just not mention "popularity" at all.

blondered: In this PR. And you need to do this only for your new module. In other places we still have "popularity" and it is correct. We are changing the naming in docstrings for de-biasing in order to not mix train dataset "popularity" and test interactions "number of users that interacted with item" which directly affects all metrics computation.
We're also very open to discuss other formulations of this concept :)

rectools/metrics/classification.py
blondered: Right

blondered: And the method remains public. So this is actually a breaking change and we need to write it in the changelog.

feldlime: Why is it a breaking change? If you don't specify the debias_config, your old code will continue to work, no?

blondered: We're adding a required (not optional) argument to a public method. So old code will not work if someone actually used this method.
To make it non-breaking we need this argument to be optional and treat `None` as `False`

feldlime: The argument is not optional, but it has a default value. The old code should work :)

blondered: You are right. My bad :)

rectools/metrics/debias.py
In48semenov: ok, then I don't understand how to use the debounce mechanism in cases where metrics are counted with `merge_reco` output. For example in these places: [calc_ranking_metrics](https://github.com/MobileTeleSystems/RecTools/blob/main/rectools/metrics/ranking.py#L534) or [calc_from_merged method from Ranking metrics](https://github.com/MobileTeleSystems/RecTools/blob/main/rectools/metrics/ranking.py#L373)

rectools/metrics/debias.py
In48semenov: Fixed. Is it correct now?

tests/metrics/test_debias.py
In48semenov: **Done**

rectools/metrics/classification.py
In48semenov: Is it correct now?

rectools/metrics/ranking.py
In48semenov: I added `DibiasableMetrikAtK` to `debias.py`

rectools/metrics/classification.py
In48semenov: Am I understanding you correctly that we come to the point where we add an argument `is_confusion_df_debiased` to `ClassificationMetric`, `ClassificationMetric`? 
I added `is_confusion_df_debiased` as argument, because the `calc` method reaches the `calc_per_user_from_confusion_df` method, where this argument is needed. But if you don't pass it to `calc` and use debias mechanism, you will get an error with this [algorithm](https://github.com/MobileTeleSystems/RecTools/pull/152#discussion_r1633793551)

blondered: We need `is_confusion_df_debiased` as an argument to `calc_per_user_from_confusion_df` method. Not as an argument to `ClassificationMetric` class.
We set it to `False` by default.
If we did debiasing in the code then we call `calc_per_user_from_confusion_df` method with `is_confusion_df_debiased` set to True.

blondered: Yes, it's correct now, thank you!
@feldlime will resolve since it is his thread

rectools/metrics/debias.py
```diff
@@ -0,0 +1,125 @@
...
+
+    debias_config: DebiasConfig = attr.ib(default=None)
+
+    def make_debias(self, interactions: pd.DataFrame) -> pd.DataFrame:
+        """
+        Downsample the size of interactions, excluding some interactions with popular items.
+
+        Algorithm:
+
+            1. Calculate item popularity distribution from interactions;
```
blondered, line 68:
```suggestion
            1. Calculate item "popularity" (here: number of unique users that had interaction with the item) distribution from interactions;
```

rectools/metrics/debias.py
```diff
@@ -0,0 +1,125 @@
...
+            2. Find first (Q1) and third (Q3) quartiles in items popularity distribution;
```
blondered, line 69:
```suggestion
            2. Find first (Q1) and third (Q3) quartiles in items "popularity" distribution;
```

blondered (changes requested):

rectools/metrics/debias.py
blondered: Yes, it is correct in the code, thank you! I proposed some changes to the docstrings below.
@feldlime will resolve this when he is happy :)

rectools/metrics/classification.py
```diff
@@ -44,8 +45,15 @@ class ClassificationMetric(MetricAtK):
     ----------
     k : int
         Number of items at the top of recommendations list that will be used to calculate metric.
+    debias_config : DebiasConfig, default None
+        Config with debias method parameters (iqr_coef, random_state).
+    is_confusion_df_debiased : bool, default False
```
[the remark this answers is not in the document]
In48semenov: **Done**

rectools/metrics/classification.py
```diff
@@ -44,8 +45,15 @@ class ClassificationMetric(MetricAtK):
...
+        If ``True``, indicator that a debias mechanism was applied before the confusion_df was computed.
+        If ``False``, indicator that no debias mechanism was applied before the confusion_df was computed.
     """
 
+    is_confusion_df_debiased: bool = attr.ib(default=False)
```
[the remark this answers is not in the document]
In48semenov: **Done**

rectools/metrics/classification.py
feldlime: yes

rectools/metrics/debias.py
feldlime: Seems we have a misunderstanding here :)

I'll try to rephrase: what will break if you remove this `if` and the code under it?

In48semenov: Yes, I see your point now :)
Fixed

rectools/metrics/debias.py
```diff
@@ -0,0 +1,125 @@
...
+    random_state : float, default None
+        Pseudorandom number generator state to control the down-sampling.
+    """
+
+    iqr_coef: float = attr.ib(default=1.5)
+    random_state: tp.Optional[int] = attr.ib(default=None)
+
+
+@attr.s
+class DibiasableMetrikAtK(MetricAtK):
```
blondered, line 45:
```suggestion
class DebiasableMetrikAtK(MetricAtK):
```

rectools/metrics/classification.py
```diff
@@ -33,7 +34,7 @@
 
 
 @attr.s
-class ClassificationMetric(MetricAtK):
+class ClassificationMetric(DibiasableMetrikAtK):
```
blondered, line 37: We need to add `DibiasableMetrikAtK` inheritange to `PartialAUC` and `PAP` metrics. We had a PR with them merged recently.

blondered (changes requested):

rectools/metrics/classification.py
In48semenov: I added `DibiasableMetrikAtK` inheritance from `PartialAUC` and `PAP`.
I have the following questions: Is it correct to apply debias method for `outer_merged_enriched` which is returned after `AUCMetric.fit(...)`? This is main point, since the calculation of `PartialAUC` and `PAP` metrics actually happens in the `calc_per_user_from_fitted` method.
If this is not correct, then I suggest applying debias method before `AUCMetric.fit(...)`, for example in the `calc_per_user` method. If the user calls `calc_from_fitted` or `calc_per_user_from_fitted `and passes `DebiasConfig`, then throw an exception.
@blondered @feldlime WDYT?

blondered: `outer_merged` has a very strict and complicated structure. Simple downsampling doesn't work, there should be all ranks present in the table. When test positive needs to be removed we need to change "__test_positive" to False if rank is not Nan. Or delete the row if rank is None. This is all not obvious.
`outer_merged` in `AucFitted` instance attribute is enriched and even more complicated. We shouldn't downsample it from there. It's a pain and it's not optimal anyway.

I believe the best way will be to:
- iterate between downsampling configs in `calc_auc_metrics`
- for each config downsample interactions and then run `fit`
- for each metric calculate results from a specific fitted instance of `_AUCMetric`
It's quite similar to what you did for `MAP`

- Also we need to add downsampling in `calc_per_user` method of `_AUCMetric` right before fitting.
- And add `is_debiased` argument to `calc_from_fitted` and `calc_per_user_from_fitted`. We need to think about the naming so that we have one name for both AUC and classification metrics. I think one name will be better. So we can do smth like `is_data_debiased` or `is_input_debiased` or simple `is_debiased` everywhere.
- Remove `is_interactions_debiased` from `calc` and `calc_per_user`

@feldlime WDYT?

feldlime: Sounds fine, however I'm not really into it.  
Please cover all possible cases with the test (direct metric calculation, using `calc_metrics` with multiple metrics and same debias config, using it with different debias configs)

In48semenov: Fixed this point.

rectools/metrics/classification.py
```diff
@@ -124,12 +137,24 @@ def calc_per_user_from_confusion_df(self, confusion_df: pd.DataFrame, catalog: C
...
+            If ``False``, indicator that no debias mechanism was applied before the confusion_df was computed.
 
         Returns
         -------
         pd.Series
             Values of metric (index - user id, values - metric value for every user).
         """
+        if not is_debiased and self.debias_config is not None:
+            raise ValueError(
+                "You have specified `debias_config` for metric but `confusin_df` is not assumed to be de-biased. "
```
blondered, line 151:
```suggestion
                "You have specified `debias_config` for metric and tried to calculate if from `confusion_df` but `confusin_df` is not assumed to be de-biased. "
```

rectools/metrics/classification.py
```diff
@@ -222,12 +257,24 @@ def calc_per_user_from_confusion_df(self, confusion_df: pd.DataFrame) -> pd.Seri
...
```
blondered, line 271: Same comments as for `ClassificationMetric`

tests/metrics/test_classification.py
```diff
@@ -38,12 +45,14 @@
...
 CATALOG = list(range(10))
 EMPTY_INTERACTIONS = pd.DataFrame(columns=[Columns.User, Columns.Item], dtype=int)
+DEBIAS_CONFIG = DebiasConfig(iqr_coef=1.5, random_state=32)
 
 
 class TestPrecision:
     def setup_method(self) -> None:
         self.metric = Precision(k=2)
         self.r_precision = Precision(k=2, r_precision=True)
+        self.metric_debias = Precision(k=2, debias_config=DEBIAS_CONFIG)
```
blondered, line 55: I don't see where you use this

In48semenov: **Done**

rectools/metrics/auc.py
```diff
@@ -447,22 +479,35 @@ def _get_sufficient_reco_explanation(self) -> str:
...
+            If ``False``, indicator that the debias mechanism has not been applied before
+            and then it will be applied if it was needed.
 
         Returns
         -------
         pd.Series
             Values of metric (index - user id, values - metric value for every user).
         """
-        outer_merged = fitted.outer_merged_enriched
+        if not is_debiased and self.debias_config is not None:
```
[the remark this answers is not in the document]
In48semenov: **Done**
Since similar functionality is used in several metric classes (auc and classification), I made a generic `_calc_debias` function in `DebiasableMetricAtK.

rectools/metrics/debias.py
```diff
@@ -0,0 +1,166 @@
...
+            Downsampling interactions.
+        """
+        if len(interactions) == 0:
+            return interactions
+
+        interactions_for_debiasing = interactions.copy()
+
+        num_users_interacted_with_item = interactions_for_debiasing.drop_duplicates(subset=Columns.UserItem)[
+            Columns.Item
+        ].value_counts()
```
feldlime, line 109: A bit shorter, faster and more memory efficient
```suggestion
        num_users_interacted_with_item = interactions_for_debiasing.groupby(Columns.Item, sort=False)[Columns.User].nunique()
```

rectools/metrics/debias.py
```diff
@@ -0,0 +1,166 @@
...
+
+        result_dfs = [interactions_result, interactions_downsampling]
+        interactions_result = pd.concat(result_dfs, ignore_index=True)
+
+        return interactions_result
+
+
+def calc_debias_for_fit_metrics(
+    metrics: tp.Dict[str, DebiasableMetrikAtK], interactions: pd.DataFrame
+) -> tp.Dict[DebiasConfig, tp.List[tp.Union[int, pd.DataFrame]]]:
```
feldlime, line 136: If you need to put exactly 2 objects, especially with different types, you should definitely use tuple, never list
```suggestion
) -> tp.Dict[DebiasConfig, tp.Tuple[int, pd.DataFrame]]:
```

rectools/metrics/debias.py
```diff
@@ -0,0 +1,166 @@
...
+            .head(max_border)
+        )
+
+        result_dfs = [interactions_result, interactions_downsampling]
+        interactions_result = pd.concat(result_dfs, ignore_index=True)
+
+        return interactions_result
+
+
+def calc_debias_for_fit_metrics(
```
feldlime, line 134: 1. Let's replace `calc_debias` to smth more informative. You're not calculating debiase here actually
2. Let's remove `for_fit_metrics`, it can be used anywhere. Call functions according with what they do, do not restrict their area of usage if not nessesary

rectools/metrics/debias.py
```diff
@@ -0,0 +1,166 @@
...
+    metrics: tp.Dict[str, DebiasableMetrikAtK], interactions: pd.DataFrame
```
feldlime, line 135: Function should accept only what it needs. Here you don't use the keys from the dictionary, hence you don't need the dictionary. Let's replace it with `tp.Iterable[DebiasableMetrikAtK]`

feldlime (changes requested):

rectools/metrics/debias.py
In48semenov: I used a sheet because the values within it could change as you go through the metrics.
I added functionality: transforming tuple into list -> changing values in it -> transforming list into tuple.

blondered: Let's talk about the whole method again. How about:
1. return `Tuple[DebiasConfig, int, pd.DataFrame]` where int is for max_k and pd.DataFrame is for debiased interactions. we don't need a dict right now so we can simplify it here
2. Let's call this `calc_debiased_fit_task`. This is used only for metrics with `fit` method because we need max_k only for `fit` and nowhere else.
3. Let's do smth like:
```
configs = set([metric.debias_config for metric in metrics])
conf_to_k = {config: [] for config in configs}
for metric in metrics:
    conf_to_k[metric.config].append(metric.k)
max_k_for_config = {config: max(k_requests) for config, k_requests in conf_to_k.items()}
res = ((config, max_k_for_config[config], DebiasableMetricAtK.make_debias(interactions, config) for config in configs)
return res
```
Here we make `make_debias` a classmethod which accepts debias config as an argument. And it is actually quite useful for external users. We can even make it a separate method which is even more useful. I would just name it `debias_interactions`

rectools/metrics/auc.py
```diff
@@ -513,12 +535,28 @@ def calc_auc_metrics(
     """
     results = {}
 
-    k_max = max(metric.k for metric in metrics.values())
     insufficient_handling_needed = any(
-        metric.insufficient_handling != InsufficientHandling.IGNORE for metric in metrics.values()
+        metric.insufficient_handling != InsufficientHandling.IGNORE
+        for metric in metrics.values()
+        if metric.debias_config is None
```
feldlime, line 541: Why do we check only metrics without debias, but then use `insufficient_handling_needed` for all metrics, including that ones with debias?

rectools/metrics/debias.py
```diff
@@ -0,0 +1,162 @@
...
+        Config with debias method parameters (iqr_coef, random_state).
+    """
+
+    debias_config: DebiasConfig = attr.ib(default=None)
+
+    def _check_debias(self, is_debiased: bool, type_of_intermediate_values_of_interactions: str) -> None:
+        if not is_debiased and self.debias_config is not None:
+            raise ValueError(
+                "You have specified `debias_config` for metric "
+                f"but `{type_of_intermediate_values_of_interactions}` is not assumed to be de-biased. "
```
feldlime, line 66: Current version sounds like "this thing should not be debiased"
```suggestion
                f"but `{type_of_intermediate_values_of_interactions}` is not de-biased. "
```

rectools/metrics/debias.py
```diff
@@ -0,0 +1,166 @@
...
+    ----------
+    k : int
+        Number of items at the top of recommendations list that will be used to calculate metric.
+    debias_config : DebiasConfig, default None
+        Config with debias method parameters (iqr_coef, random_state).
+    """
+
+    debias_config: DebiasConfig = attr.ib(default=None)
+
+    def _check_debias(self, is_debiased: bool, name_data: str) -> None:
```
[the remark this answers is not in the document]
feldlime: No, please, let's make it shorter :) `obj_name` is enough

rectools/metrics/debias.py
```diff
@@ -0,0 +1,166 @@
...
+        if not is_debiased and self.debias_config is not None:
+            raise ValueError(
+                "You have specified `debias_config` for metric "
+                f"but `{name_data}` is not assumed to be de-biased. "
+                f"Please make de-biasing for `{name_data}` "
+                "and specify `is_debiased` as `True` "
+                "or otherwise use `calc` and `calc_per_user` methods for auto de-biasing."
+            )
+
+    def make_debias(self, interactions: pd.DataFrame) -> pd.DataFrame:
```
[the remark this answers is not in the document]
feldlime: Sorry for confusion.  
I meant to rename function, not parameter

rectools/metrics/debias.py
feldlime: > we don't need a dict right now so we can simplify it here

Having a dict is good to show that we don't have 2 items with the same configs

> `set([x for x in xx])` 

`set(x for x in xx)` 

> `k_requests`

Sounds like we are counting requests. Let's keep just `k` or `requested_k`

Agreed with the rest

blondered (changes requested):

tests/metrics/test_auc.py
```diff
@@ -151,6 +153,81 @@ def test_when_duplicates_in_interactions_insufficient(
...
+        "k, insufficient_handling",
+        (
+            (1, InsufficientHandling.IGNORE),
+            (3, InsufficientHandling.IGNORE),
+            (1, InsufficientHandling.EXCLUDE),
+            (3, InsufficientHandling.EXCLUDE),  # user 2 was excluded
+        ),
+    )
+    def test_calc_debias(self, k: int, insufficient_handling: str) -> None:
+        reco = pd.DataFrame(
```
blondered, line 166: Let's do the same structure for test classes and methods as you did for ranking module and stop duplicating a lot of code

tests/metrics/test_classification.py
```diff
@@ -195,3 +203,83 @@ def test_when_no_interactions(self) -> None:
             expected_metric_per_user,
         )
         assert np.isnan(self.metric.calc(RECO, EMPTY_INTERACTIONS))
+
+
+class TestDebiasMetric:
```
blondered, line 208: That's a lot fo `if`s. Seems that making separate classes TestDebiasableClassificationMetric and TestDebiasableSimpleClassificationMetric would have solved the problem

blondered (changes requested):

tests/metrics/test_debias.py
```diff
@@ -0,0 +1,130 @@
...
+        k_max_debias = calc_debiased_fit_task(metrics=metrics_fitted.values(), interactions=interactions)
+
+        unique_debias_config = []
+        for metric in metrics_fitted.values():
+            if metric.debias_config not in unique_debias_config:
+                unique_debias_config.append(metric.debias_config)
+
+        expected_result = len(unique_debias_config)
+
+        assert len(k_max_debias) == expected_result
```
blondered, line 130: Let's add max_k correct values check at least

rectools/metrics/ranking.py
```diff
@@ -565,9 +592,21 @@ def calc_ranking_metrics(
```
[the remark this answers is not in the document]
feldlime: Not sure. Why do we do this?

rectools/metrics/debias.py
```diff
@@ -0,0 +1,206 @@
...
+
+    Returns
+    -------
+    dict(DebiasConfig->list[(int | pd.DataFrame)])
+        Dictionary, where key is debias config
+        and values are a tuple of the corresponding `k_max` and de-basing `interactions`.
+    """
+    debiasing_interactions = calc_debiased_different_configs(metrics, interactions, debiasing_interactions_prev)
+
+    max_k_for_config = {config: 0 for config in debiasing_interactions}
```
feldlime, line 160:
```suggestion
    max_k_for_config = defaultdict(int)
```

rectools/metrics/debias.py
```diff
@@ -0,0 +1,206 @@
...
+    -------
+    dict(DebiasConfig->pd.DataFrame])
+        Dictionary, where key is debias config and values are de-basing `interactions`.
+    """
+    if debiasing_interactions_prev is not None:
+        configs_new = set(
+            metric.debias_config for metric in metrics if metric.debias_config not in debiasing_interactions_prev
+        )
+    else:
+        configs_new = set(metric.debias_config for metric in metrics)
```
feldlime, line 197:
```suggestion
    configs_new = set(metric.debias_config for metric in metrics)
    if debiasing_interactions_prev is not None:
        configs_new -= set(debiasing_interactions_prev.keys())
```

rectools/metrics/debias.py
```diff
@@ -0,0 +1,206 @@
...
+    result = {
+        config: (max_k_for_config[config], d_interaciotns) for config, d_interaciotns in debiasing_interactions.items()
+    }
+    return result
+
+
+def calc_debiased_different_configs(
+    metrics: tp.Iterable[DebiasableMetrikAtK],
+    interactions: pd.DataFrame,
+    debiasing_interactions_prev: tp.Optional[tp.Dict[DebiasConfig, pd.DataFrame]] = None,
```
feldlime, line 173: Are you sure it's `debiasing`, not `debiased`? @blondered help

blondered: `prev_debiased_interactions` here and everywhere

blondered (changes requested):

rectools/metrics/debias.py
```diff
@@ -0,0 +1,206 @@
...
+        .head(max_border)
+    )
+
+    result_dfs = [interactions_result, interactions_downsampling]
+    interactions_result = pd.concat(result_dfs, ignore_index=True)
+
+    return interactions_result
+
+
+def calc_debiased_fit_task(
```
blondered, line 133:
```suggestion
def debias_for_metric_configs(
```

rectools/metrics/debias.py
```diff
@@ -0,0 +1,206 @@
...
+        Dictionary, where key is debias config and values are de-basing `interactions`.
+    """
+    if debiasing_interactions_prev is not None:
+        configs_new = set(
+            metric.debias_config for metric in metrics if metric.debias_config not in debiasing_interactions_prev
+        )
+    else:
+        configs_new = set(metric.debias_config for metric in metrics)
+
+    debiasing_interactions = {
```
blondered, line 199:
```suggestion
    debiased_interactions = {
```

rectools/metrics/debias.py
```diff
@@ -0,0 +1,206 @@
...
+    for metric in metrics:
+        max_k_for_config[metric.debias_config] = max(max_k_for_config[metric.debias_config], metric.k)
+
+    result = {
+        config: (max_k_for_config[config], d_interaciotns) for config, d_interaciotns in debiasing_interactions.items()
+    }
+    return result
+
+
+def calc_debiased_different_configs(
```
blondered, line 170:
```suggestion
def debias_for_metric_configs(
```

blondered (changes requested):

rectools/metrics/ranking.py
```diff
@@ -558,16 +593,24 @@ def calc_ranking_metrics(
     """
     results = {}
 
+    debiasing_merged = None
```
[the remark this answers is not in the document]
blondered: Sorry, my bad

blondered (dismissed):

CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [0.7.1] - x.x.2024
+
```
feldlime, line 12: Please remove these 2 lines, it's enough to have `Unreleased` above. We will decide the date and version during the release.

rectools/metrics/auc.py
```diff
@@ -88,6 +89,8 @@ class _AUCMetric(MetricAtK):
         until the model has non-zero scores for the item in item-item similarity matrix. So with
         small `K` for neighbours in ItemKNN and big `K` for `recommend` and AUC based metric you
         will still get an error when `insufficient_handling` is set to `raise`.
+    debias_config : DebiasConfig, default None
```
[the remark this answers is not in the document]
feldlime: You marked this as resolved, but the issue is not fixed

rectools/metrics/auc.py
```diff
@@ -217,36 +220,46 @@ def calc_per_user(self, reco: pd.DataFrame, interactions: pd.DataFrame) -> pd.Se
...
         """
         Calculate metric values for all users from from fitted data.
 
         Parameters
         ----------
         fitted : AUCFitted
             Meta data that got from `.fit` method.
+        is_debiased : bool, default False
+            If ``True``, indicator that a debias mechanism has been applied before.
+            If ``False``, indicator that the debias mechanism has not been applied before.
```
feldlime, line 262:
```suggestion
            An indicator of whether the debias transformation has been applied before or not.
```

rectools/metrics/auc.py
```diff
@@ -307,6 +320,8 @@ class PartialAUC(_AUCMetric):
         until the model has non-zero scores for the item in item-item similarity matrix. So with
         small `K` for neighbours in ItemKNN and big `K` for `recommend` and AUC based metric you
         will still get an error when `insufficient_handling` is set to `raise`.
+    debias_config : DebiasConfig, default None
```
feldlime, line 323:
```suggestion
    debias_config : DebiasConfig, optional, default None
```

rectools/metrics/auc.py
```diff
@@ -339,25 +354,27 @@ def _get_sufficient_reco_explanation(self) -> str:
...
         """
         Calculate metric values for all users from from fitted data.
 
         Parameters
         ----------
         fitted : AUCFitted
             Meta data that got from `.fit` method.
+        is_debiased : bool, default False
+            If ``True``, indicator that a debias mechanism has been applied before.
+            If ``False``, indicator that the debias mechanism has not been applied before.
```
feldlime, line 367:
```suggestion
            An indicator of whether the debias transformation has been applied before or not.
```

rectools/metrics/auc.py
feldlime: I won't repeat it anymore, please fix everywhere

rectools/metrics/auc.py
feldlime: I won't repeat it anymore, please fix everywhere

rectools/metrics/debias.py
```diff
@@ -0,0 +1,204 @@
...
+
+@attr.s(frozen=True)
+class DebiasConfig:
+    """
+    Config with debias method parameters.
+
+    Parameters
+    ----------
+    iqr_coef : float, default 1.5
+        Coefficient for defining as the maximum value inside the border.
```
feldlime, line 36: Let's please rewrite this to give more information, now it really doesn't help me to understand what is it used for.  At least let's add `interquartile range` phrase and some information of how it used for debiasing.  
Maybe also a little bit more information above to the description `DebiasConfig` itself. IMO now it's unclear what is it.

rectools/metrics/debias.py
```diff
@@ -0,0 +1,204 @@
...
+    """
+
+    iqr_coef: float = attr.ib(default=1.5)
+    random_state: tp.Optional[int] = attr.ib(default=None)
+
+
+@attr.s
+class DebiasableMetrikAtK(MetricAtK):
+    """
+    Classification metric base class.
```
feldlime, line 48: not classification

tests/metrics/test_auc.py
```diff
@@ -277,3 +279,115 @@ def test_when_duplicates_in_interactions_insufficient(
...
+                PAP(k=1, insufficient_handling=InsufficientHandling.EXCLUDE, debias_config=DEBIAS_CONFIG),
+            ),
+            (
+                PAP(k=3, insufficient_handling=InsufficientHandling.EXCLUDE),
+                PAP(k=3, insufficient_handling=InsufficientHandling.EXCLUDE, debias_config=DEBIAS_CONFIG),
+            ),
+        ),
+    )
+    def test_calc(self, metric: tp.Union[PartialAUC, PAP], metric_debias: tp.Union[PartialAUC, PAP]) -> None:
+        downsample_interactions = debias_interactions(self.interactions, config=metric_debias.debias_config)
```
feldlime, line 338:
```suggestion
        debiased_interactions = debias_interactions(self.interactions, config=metric_debias.debias_config)
```

tests/metrics/test_auc.py
```diff
@@ -277,3 +279,115 @@ def test_when_duplicates_in_interactions_insufficient(
...
+
+        expected_metric_per_user_downsample = metric.calc_per_user(self.reco, downsample_interactions)
```
feldlime, line 340:
```suggestion
        expected_metric_per_user = metric.calc_per_user(self.reco, downsample_interactions)
```

tests/metrics/test_auc.py
```diff
@@ -277,3 +279,115 @@ def test_when_duplicates_in_interactions_insufficient(
...
+        result_metric_per_user = metric_debias.calc_per_user(self.reco, self.interactions)
```
feldlime, line 341:
```suggestion
        actual_metric_per_user = metric_debias.calc_per_user(self.reco, self.interactions)
```

tests/metrics/test_auc.py
feldlime: (Everywhere)

tests/metrics/test_auc.py
feldlime: (Everywhere)

tests/metrics/test_auc.py
feldlime: (Everywhere)

tests/metrics/test_auc.py
feldlime: It may look like personal preferences, but actually it's really important to name your variables, functions, classes, etc. properly to make it easier to read and understand your code for other developers.

Here are 2 problems:
1. Grammatically incorrect name: "downsample interactions" is an imperative form, so it could be the name of function not variable. Here it should be "downsample**d** interactions".
2. Inconsistency: you call function `debias_interactions` over interactions, it's expected that in the result you should have "debiased interactions", not "downsampled". There may be exceptions from this rule, but here I don't see any reason for different naming.

tests/metrics/test_auc.py
```diff
@@ -277,3 +279,115 @@ def test_when_duplicates_in_interactions_insufficient(
...
+        result_calc = metric_debias.calc(self.reco, self.interactions)
```
feldlime, line 342:
```suggestion
        actual_metric = metric_debias.calc(self.reco, self.interactions)
```

feldlime: (Everywhere)

tests/metrics/test_auc.py
```diff
@@ -277,3 +279,115 @@ def test_when_duplicates_in_interactions_insufficient(
...
+                PAP(k=1, insufficient_handling=InsufficientHandling.EXCLUDE),
+                PAP(k=1, insufficient_handling=InsufficientHandling.EXCLUDE, debias_config=DEBIAS_CONFIG),
+            ),
+            (
+                PAP(k=3, insufficient_handling=InsufficientHandling.EXCLUDE),
+                PAP(k=3, insufficient_handling=InsufficientHandling.EXCLUDE, debias_config=DEBIAS_CONFIG),
+            ),
+        ),
+    )
+    def test_calc(self, metric: tp.Union[PartialAUC, PAP], metric_debias: tp.Union[PartialAUC, PAP]) -> None:
```
feldlime, line 337: You don't really need to duplicate for each metric it's version with and without debias config.  
You can keep just version without config and inside the function make 
`debiased_metric = PAP(k=metric.k, insufficient_handling=metric.insufficient_handling, debias_config=DEBIAS_CONFIG)`

feldlime: (Everywhere)

tests/metrics/test_auc.py
```diff
@@ -277,3 +279,115 @@ def test_when_duplicates_in_interactions_insufficient(
...
+    @pytest.mark.parametrize(
+        "metric_debias",
+        (
+            PartialAUC(k=3, insufficient_handling=InsufficientHandling.IGNORE, debias_config=DEBIAS_CONFIG),
+            PartialAUC(k=3, insufficient_handling=InsufficientHandling.EXCLUDE, debias_config=DEBIAS_CONFIG),
+            PAP(k=3, insufficient_handling=InsufficientHandling.IGNORE, debias_config=DEBIAS_CONFIG),
+            PAP(k=3, insufficient_handling=InsufficientHandling.EXCLUDE, debias_config=DEBIAS_CONFIG),
+        ),
+    )
+    def test_when_no_interactions(self, metric_debias: tp.Union[PartialAUC, PAP]) -> None:
```
feldlime, line 356:
```suggestion
    def test_when_no_interactions(self, debiased_metric: tp.Union[PartialAUC, PAP]) -> None:
```

feldlime: (Everywhere)

tests/metrics/test_debias.py
```diff
@@ -0,0 +1,230 @@
...
+                    "dPAP@3": PAP(k=3, debias_config=DEBIAS_CONFIG_DEFAULT),
+                    "dPAP@2": PAP(k=2, debias_config=DebiasConfig(iqr_coef=1.6, random_state=32)),
+                },
+            ),
+        ),
+    )
+    def test_calc_debiased_fit_task(
+        self,
+        metrics: tp.Dict[str, DebiasableMetrikAtK],
+        num_metrics_prev: tp.Optional[tp.Dict[str, DebiasableMetrikAtK]],
```
feldlime, line 139: Why it's `num` if it's a dict?

tests/metrics/test_scoring.py
```diff
@@ -166,3 +170,52 @@ def test_raises(self, metric: MetricAtK, arg_names: tp.List[str]) -> None:
...
+            "debias_hitrate@3": HitRate(k=3, debias_config=debias_config),
+            "debias_map@1": MAP(k=1, debias_config=debias_config),
+            "debias_map@3": MAP(k=3, debias_config=debias_config),
+            "debias_ndcg@3": NDCG(k=3, debias_config=debias_config),
+            "debias_mrr@3": MRR(k=3, debias_config=debias_config),
+            "debias_pap@3": PAP(k=3, debias_config=debias_config),
+            "debias_partauc@3": PartialAUC(k=3, debias_config=debias_config),
+        }
+        metrics = {
+            "debias_precision@3": Precision(k=3),
```
feldlime, line 192: Why these metrics called `debias_...` while they are not debiased?

tests/metrics/test_scoring.py
```diff
@@ -166,3 +170,52 @@ def test_raises(self, metric: MetricAtK, arg_names: tp.List[str]) -> None:
...
+            "debias_ndcg@3": NDCG(k=3),
+            "debias_mrr@3": MRR(k=3),
+            "debias_pap@3": PAP(k=3),
+            "debias_partauc@3": PartialAUC(k=3),
+        }
+
+        interactions_downsampling = debias_interactions(self.interactions, config=debias_config)
+
+        actual = calc_metrics(
+            metrics=debias_metrics,  # type: ignore
```
feldlime, line 210: Why do we need ignore here and below?

feldlime (changes requested):

tests/metrics/test_scoring.py
In48semenov: I have used the same name for consistency of dict keys, if i was to change the name it would throw an error for this test

do you have any other naming suggestions instead?

tests/metrics/test_scoring.py
In48semenov: I use this 'ignore' because i found an interesting situation, the lint gives an error when i use only metric as DebiasableMetrikAtK in dict, but i dont  give the dictionary metrics that are directly inherited by MetrikAtK

tests/metrics/test_scoring.py
feldlime: ok, let's keep it as is

tests/metrics/test_scoring.py
feldlime: Could you elaborate a bit on this? I don't understand

In48semenov: If I remove `# type: ignore`, I get the following error from `mypy` linter: 


But if I add to the current metrics dictionaries in this test that don't inherit from `DebiasableMetricAtK`, there is no error.

And since this test is aimed at testing `DebiasableMetricAtK` metrics, it won't be correct to add here metrics that don't belong to this class.

If you know how to solve this problem, I would be grateful.

feldlime: Ok, for the future: we're using linters to make the code better and avoid as many mistakes as possible. If linter gives us an error, we study it and try to fix it. If you have a problem with understanding a problem feel free to ask, we are happy to help. We ignore linter only when we are absolutely sure that the error is false positive. 
If we ignored every error we didn't understand, then it wouldn't make any sense to use linters, right?

If mypy gives you a link to a page with the problem explanation, it could be very useful to read it sometimes :)  
Here to fix the problem it's enough to add an explicit type annotation like `metrics: tp.Dict[str, MetricAtK]`, but it's better to use the hint from mypy and change `Dict` to `Mapping` in the function type annotation

In48semenov: Is it okay if I change the annotation in the [select_by_type](https://github.com/In48semenov/RecTools/blob/feature/debias_wrapper/rectools/metrics/scoring.py#L120) function for `objects` from `Dict` to `Mapping` along with this change? 
Because after changing the annotation in `calc_metrics` mypy gives the following error:
```
Argument 1 to "select_by_type" has incompatible type "Mapping[str, MetricAtK]"; expected "Dict[Any, Any]"
```

CHANGELOG.md
In48semenov: Fixed

rectools/metrics/auc.py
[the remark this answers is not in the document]
In48semenov: Sorry, my bad
Fixed

rectools/metrics/auc.py
In48semenov: Fixed

rectools/metrics/auc.py
In48semenov: Fixed

rectools/metrics/auc.py
In48semenov: Fixed

rectools/metrics/debias.py
In48semenov: Is that more understandable?

rectools/metrics/debias.py
In48semenov: Fixed

tests/metrics/test_auc.py
In48semenov: Fixed

tests/metrics/test_auc.py
In48semenov: Sorry, my bad
Fixed

tests/metrics/test_auc.py
In48semenov: Fixed

tests/metrics/test_auc.py
In48semenov: Fixed

tests/metrics/test_auc.py
In48semenov: Fixed

tests/metrics/test_auc.py
In48semenov: Fixed

tests/metrics/test_debias.py
In48semenov: Fixed

tests/metrics/test_scoring.py
feldlime: Yes, you can change it to `Mapping[Any, Any]`  
Let me know if you get any new error from mypy

rectools/metrics/debias.py
feldlime: Yes, perfect, thank you!

feldlime (approved):

Issue #150, opened 2024-05-28
Add `model_meta` processing to `MetricsApp` for labels colouring


blondered:
### Feature Description

- Add optional argument `model_meta` to `MetricsApp.construct` method
- `model_meta` is pd.DataFrame with obligatory `Columns.Model` columns. All other are model params
- Join models meta and models metrics during `construct` method
- In the widget add option to colour dots on scatter plot not by model name but by any of the provided model meta params (like dimension size, k_neighbours, etc.)

### Why this feature?

It is useful for real services research

PR #154, opened 2024-06-19
Feature/metricsapp metainfo

Gooogr:
## Description
Closes https://github.com/MobileTeleSystems/RecTools/issues/150: add support of scatter plot colors maps based on any model meta-info.

## Type of change

- [ ] Optimization

## How Has This Been Tested?


Gooogr:
TODO:
* Update docstring with meta-info example
* Expand tests
* Update .ipynb example (cross-validation)

Gooogr:
TODO:
 - [ ] Improve test coverage for codecov

rectools/visuals/metrics_app.py
```diff
@@ -134,13 +165,54 @@ def _validate_input_data(self) -> None:
...
+            raise ValueError("Found ambiguous values in `Model` column of `models_metadata`")
+
     @lru_cache
-    def _make_chart_data(self, fold_number: int) -> pd.DataFrame:  # pragma: no cover
-        return self.models_metrics[self.models_metrics[Columns.Split] == fold_number].drop(columns=Columns.Split)
+    def _make_chart_data(self, fold_number: int) -> pd.DataFrame:
+        return (
+            self.models_metrics[self.models_metrics[Columns.Split] == fold_number]
+            .drop(columns=Columns.Split)
+            .merge(self.models_metadata, on=Columns.Model, how="left")
```
blondered, line 181: I have a feeling that it will be more clear to merge `models_metadata` with `models_metrics` one time during `contruct` and also remember col names for metrics and meta. And then accept merged data and col names in __init__. WDYT?
@feldlime WDYT?

rectools/visuals/metrics_app.py
```diff
@@ -124,7 +155,7 @@ def fold_ids(self) -> tp.List[int]:
         """Sorted list of fold identifiers from the `models_metrics`."""
         return sorted(self.models_metrics[Columns.Split].unique())
 
-    def _validate_input_data(self) -> None:
+    def _validate_models_metrics(self) -> None:
         if not isinstance(self.models_metrics, pd.DataFrame):
             raise ValueError("Incorrect input type. `metrics_data` should be a DataFrame")
         if Columns.Split not in self.models_metrics.columns:
```
blondered, line 161: Can we add a possibility for user to pass models_metrics without a split column? Assuming that there is only one fold.
Can we do it in this PR or should we do it afterwards?

Gooogr: Nice idea, should be easy to implement. I think we could add in in this PR.

rectools/visuals/metrics_app.py
Gooogr: Keeping in mind that we also want to visualize the data for [the case when we don't have a folds column](https://github.com/MobileTeleSystems/RecTools/pull/154#discussion_r1649685516), idea to make some main table before finalizing sounds nice.
But is it ok to have different arguments for `construct` and `init` methods? Seems a little bit confusing for me. But I can create additional method for data merging and cache the result. Then grab it in three other methods: for a averaged data, for a fold data and a data without any folds.

blondered: It is definetely ok to have different argument for `construct` and `init`, it is very reasonable and common. So you can merge during `construct` and pass merged data to `init`

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
```diff
@@ -85,62 +96,160 @@ def construct(
...
+        color: str,
+        legend_title: str,
+    ) -> go.Figure:
+        scatter_kwargs = {
+            "width": WIDGET_WIDTH,
+            "height": WIDGET_HEIGHT,
+        }
+        scatter_kwargs.update(self.scatter_kwargs)
+
+        data = data.sort_values(by=color, ascending=False)
```
blondered, line 235: How about changing it go `ascending=True`?
And I wrote to you about recolouring `nan` to grey separately

rectools/visuals/metrics_app.py
```diff
@@ -69,12 +74,18 @@ def construct(
...
+                - `Columns.Models` - model names
+                - Any other columns with additional information
         show_legend : bool, default True
             Specifies whether to display the chart legend.
         auto_display : bool, default True
             Automatically displays the widgets immediately after initialization.
-        layout_kwargs : tp.Optional[tp.Dict[str, tp.Any]], optional, default None
-            Additional arguments from `plotly.graph_objects.Layout`
+        scatter_kwargs : tp.Optional[tp.Dict[str, tp.Any]], optional, default None
+            Additional arguments from `plotly.express.scatter`
```
blondered, line 88:
```suggestion
            Additional arguments for `plotly.express.scatter`
```

rectools/visuals/metrics_app.py
```diff
@@ -85,62 +96,160 @@ def construct(
...
+        # get the first group's fold counts and fold names for reference
+        ref_group = grouped.get_group(next(iter(grouped.groups.keys())))
+        ref_fold_count = ref_group[Columns.Split].nunique()
+        ref_fold_names = set(ref_group[Columns.Split].unique())
+
+        for model, group in grouped:
+            if group["i_split"].nunique() != ref_fold_count:
+                raise ValueError(f"{model} does not have the expected fold amount")
+            if set(group["i_split"].unique()) != ref_fold_names:
+                raise ValueError(f"{model} does not have the expected fold names")
```
blondered, line 182:
```suggestion
        ref_group_name = next(iter(grouped.groups.keys()))
        ref_group = grouped.get_group(ref_group_name)
        ref_fold_count = ref_group[Columns.Split].nunique()
        ref_fold_names = set(ref_group[Columns.Split].unique())

        for model, group in grouped:
            if group["i_split"].nunique() != ref_fold_count:
                raise ValueError(f"""{model} does not have the same fold amount as {ref_group_name}")
            if set(group["i_split"].unique()) != ref_fold_names:
                raise ValueError(f"{model} does not have the same fold names as {ref_group_name}")
```

rectools/visuals/metrics_app.py
```diff
@@ -85,62 +96,160 @@ def construct(
...
-        return sorted(self.models_metrics[Columns.Split].unique())
+        if Columns.Split in self.data.columns:
+            return sorted(self.data[Columns.Split].unique())
+        return None
 
-    def _validate_input_data(self) -> None:
-        if not isinstance(self.models_metrics, pd.DataFrame):
+    @staticmethod
+    def _validate_models_metrics_base(models_metrics: pd.DataFrame) -> None:
+        if not isinstance(models_metrics, pd.DataFrame):
```
blondered, line 158: We can drop this check because we have type annotation that `models_metics` is a `pd.DataFrame`

rectools/visuals/metrics_app.py
```diff
@@ -85,62 +96,160 @@ def construct(
...
-        if Columns.Split not in self.models_metrics.columns:
-            raise KeyError("Missing `Split` column in `metrics_data` DataFrame")
-        if Columns.Model not in self.models_metrics.columns:
-            raise KeyError("Missing `Model`` column in `metrics_data` DataFrame")
-        if len(self.models_metrics.columns) < 3:
+        if Columns.Model not in models_metrics.columns:
+            raise KeyError("Missing `Model` column in `metrics_data` DataFrame")
+        if len([item for item in models_metrics.columns if item not in {Columns.Model, Columns.Split}]) < 1:
             raise KeyError("`metrics_data` DataFrame assumed to have at least one metric column")
+        if models_metrics.isnull().values.any():
```
blondered, line 164: Do we need this for any case apart from dq check? Will nan values break anything from our functionality?
If they don't break anything, I think we can drop this check. From what I have tried, model with nan metrics remains in the legend but doesn't show on the chart if metric with nan value is selected. Which is good I think. What do you think?
(We shill need to check null values in Model and Split columns)

rectools/visuals/metrics_app.py
```diff
@@ -85,62 +96,160 @@ def construct(
...
+            if models_names_comb.nunique() != len(models_names_comb):
+                raise ValueError("Each `Model` value in the `metrics_data` DataFrame must be unique")
+        else:
+            models_metrics[Columns.Model] = models_metrics[Columns.Model].astype(str).str.replace(" ", "_")
+            if models_metrics[Columns.Model].nunique() != len(models_metrics[Columns.Model]):
+                raise ValueError("Each `Model` value in the `metrics_data` DataFrame must be unique")
+
+    @staticmethod
+    def _validate_models_metadata(models_metadata: pd.DataFrame) -> None:
+        if not isinstance(models_metadata, pd.DataFrame):
```
blondered, line 199: We can drop this

rectools/visuals/metrics_app.py
```diff
@@ -85,62 +96,160 @@ def construct(
...
     def _make_chart_data_avg(self) -> pd.DataFrame:
-        return self.models_metrics.drop(columns=Columns.Split).groupby(Columns.Model, sort=False).mean().reset_index()
+        metric_data_columns = [Columns.Model] + self.metric_names
+        meta_data_columns = [Columns.Model] + self.meta_names
+        # separate metric data because meta data could contain strings
+        metrics_data = self.data[metric_data_columns].groupby([Columns.Model], sort=False).mean().reset_index()
+        meta_data = self.data[meta_data_columns].drop_duplicates()
+        return metrics_data.merge(meta_data, on=Columns.Model, how="left").reset_index(drop=True)
+
+    def _create_chart(
```
blondered, line 221: You are using `_create_chart` and `_update_chart` methods like they are working with the same thing. But they are actually working with `go.Figure` and `go.FigureWidget`. Can we please name them accordingly?
`_create_chart_figure` and `_update_figure_widget`

rectools/visuals/metrics_app.py
```diff
@@ -149,61 +258,106 @@ def _update_chart(
...
-        existing_traces = {trace.name: trace for trace in fig_widget.data}
+        meta_feature: widgets.Dropdown,
+        use_meta: widgets.Checkbox,
+    ) -> None:  # pragma: no cover
+        chart_data = self._create_chart_data(use_avg, fold_i)
+        color_clmn = meta_feature.value if use_meta.value else Columns.Model
+
+        # Save dots symbols from the previous widget state
+        # `split(" ", 1)[-1]` removed metainfo from trace name. Thus we guarantee to map with traces from previous state
+        trace_name2symbol = {trace.name.split(" ", 1)[-1]: trace.marker.symbol for trace in self.fig.data}
```
blondered, line 269: About the splitting:
1. Trace name starts from param name. And we didn't replace `" "` to `"_"` for parameters. So we might have a problem
2. We don't need to replace `" "` to `"_"` at all (I think). We need to replace `","` to `"_"` or something else. Or replace `", "` to `",_"`. And after that we need to split by `", "` which will solve our problem
3. Also please move this to a separate func so that you write this logic only one time, not three time in the code

rectools/visuals/metrics_app.py
```diff
@@ -149,61 +258,106 @@ def _update_chart(
...
-                        mode="markers",
-                        name=model,
-                        hovertemplate=hover,
-                        showlegend=self.show_legend,
-                    )
-            fig_widget.layout.xaxis.title = metric_x.value
-            fig_widget.layout.yaxis.title = metric_y.value
-        self.fig = go.Figure(data=fig_widget.data, layout=fig_widget.layout)
-        self.fig.layout.margin = None
+            for trace in self.fig.data:
```
blondered, line 280: Please double check me. But I think that you can just do enumerate here. Because you were enumerating `self.fig.data` to create `nometa_trace_name2idx`. And now you are iterating through the same self.fig.data. We can drop `nometa_trace_name2idx` and just enumerate here to get `idx`.

Or is there something that I am missing? Then please write a comment for the logic. 
I'm not familiar with plotly figure widgets

rectools/visuals/metrics_app.py
Gooogr: 1. We replaced `"  "` to `"_"` in line [135](https://github.com/MobileTeleSystems/RecTools/blob/38ef219326826ffd64be8acba54c124cceec973a/rectools/visuals/metrics_app.py#L135). 
2. In this case we have to check if user activate "color by meta" mode or not. Otherwise we could trim model name.
3. Sure, good point

rectools/visuals/metrics_app.py
Gooogr: No, unfortunately, it is more complicated.
The chart changes trace orders for "meta" and "no-meta" modes, which is evident in the legend sorting section. Therefore, I cannot simply iterate over them. Additionally, the meta parameter is now part of the trace name, necessitating its removal to match trace names between the "meta" and "no-meta" states. As a result, I have implemented this key mapping approach.

rectools/visuals/metrics_app.py
blondered: 1. We replaced them for model names but not for meta param names which are first in trace.name when "color by meta" is on. How about splitting `trace.name.split(", ")[-1]`? This way we don't care what's in meta param names, only model names need to be processed
2.I think it's not obligatory if we process and split correctly

rectools/visuals/metrics_app.py
Gooogr: Actually, no, we can safely keep these NaN values for metrics. I need to check how pandas handles these NaNs when computing average fold values, but as far as I can see, this is the only potential issue.

rectools/visuals/metrics_app.py
Gooogr: Updated

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done, but current recolouring is black

rectools/visuals/metrics_app.py
```diff
@@ -218,21 +217,30 @@ def _make_chart_data_avg(self) -> pd.DataFrame:
         meta_data = self.data[meta_data_columns].drop_duplicates()
         return metrics_data.merge(meta_data, on=Columns.Model, how="left").reset_index(drop=True)
 
-    def _create_chart(
+    @staticmethod
+    @lru_cache
+    def _trim_metadata(raw_string: str, splitter: str = ", ") -> tp.Tuple[str, str]:
```
blondered, line 222: Let's call it `_split_to_meta_and_model`

rectools/visuals/metrics_app.py
```diff
@@ -265,21 +269,19 @@ def _update_chart(
         color_clmn = meta_feature.value if use_meta.value else Columns.Model
 
         # Save dots symbols from the previous widget state
-        # `split(" ", 1)[-1]` removed metainfo from trace name. Thus we guarantee to map with traces from previous state
-        trace_name2symbol = {trace.name.split(" ", 1)[-1]: trace.marker.symbol for trace in self.fig.data}
+        # Remove metainfo from trace name. Thus we guarantee to map with traces from previous state
+        trace_name2symbol = {self._trim_metadata(trace.name)[1]: trace.marker.symbol for trace in self.fig.data}
```
blondered, line 273: Let's call it `model_name2symbol` since it's not full name

blondered (changes requested):

rectools/visuals/metrics_app.py
```diff
@@ -155,55 +155,53 @@ def fold_ids(self) -> tp.Optional[tp.List[int]]:
...
-                raise ValueError(f"{model} does not have the expected fold amount")
+                raise ValueError(f"""{model} does not have the same fold amount as {ref_group_name}""")
             if set(group["i_split"].unique()) != ref_fold_names:
-                raise ValueError(f"{model} does not have the expected fold names")
+                raise ValueError(f"{model} does not have the same fold names as {ref_group_name}")
 
     @staticmethod
     def _validate_models_metrics_names(models_metrics: pd.DataFrame) -> None:
         # Validate that all Models names are unique
+        models_metrics = models_metrics.copy()
```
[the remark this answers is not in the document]
Gooogr: Good catch, thank you. We no longer require this code. It originates from an older implementation where we replaced spaces with underscores.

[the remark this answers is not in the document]
Gooogr: Fixed

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
```diff
@@ -265,21 +269,19 @@ def _update_chart(
...
-        trace_name2symbol = {trace.name.split(" ", 1)[-1]: trace.marker.symbol for trace in self.fig.data}
+        # Remove metainfo from trace name. Thus we guarantee to map with traces from previous state
+        trace_name2symbol = {self._trim_metadata(trace.name)[1]: trace.marker.symbol for trace in self.fig.data}
         legend_title = f"{meta_feature.value}, {DEFAULT_LEGEND_TITLE}" if use_meta.value else DEFAULT_LEGEND_TITLE
-        self.fig = self._create_chart(chart_data, metric_x.value, metric_y.value, color_clmn, legend_title)
+        self.fig = self._create_chart_figure(chart_data, metric_x.value, metric_y.value, color_clmn, legend_title)
 
         for trace in self.fig.data:
-            trace_name = trace.name.split(" ", 1)[-1]
+            trace_name = self._trim_metadata(trace.name)[1]
```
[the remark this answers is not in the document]
Gooogr: Done

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,274 @@ def construct(
...
+            return sorted(self.data[Columns.Split].unique())
+        return None
+
+    @staticmethod
+    def _validate_models_metrics_base(models_metrics: pd.DataFrame) -> None:
+        if Columns.Model not in models_metrics.columns:
+            raise KeyError("Missing `Model` column in `metrics_data` DataFrame")
+        if len([item for item in models_metrics.columns if item not in {Columns.Model, Columns.Split}]) < 1:
             raise KeyError("`metrics_data` DataFrame assumed to have at least one metric column")
+        if models_metrics[Columns.Model].isnull().values.any():
```
feldlime, line 163:
```suggestion
        if models_metrics[Columns.Model].isnull().any():
```

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,274 @@ def construct(
...
+            raise ValueError("Found NaN values in `Model`column of `metrics_data`")
+        if Columns.Split in models_metrics.columns and models_metrics[Columns.Split].isnull().values.any():
```
feldlime, line 165:
```suggestion
        if Columns.Split in models_metrics.columns and models_metrics[Columns.Split].isnull().any():
```

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,274 @@ def construct(
...
+        return None
+
+    @staticmethod
+    def _validate_models_metrics_base(models_metrics: pd.DataFrame) -> None:
+        if Columns.Model not in models_metrics.columns:
+            raise KeyError("Missing `Model` column in `metrics_data` DataFrame")
+        if len([item for item in models_metrics.columns if item not in {Columns.Model, Columns.Split}]) < 1:
             raise KeyError("`metrics_data` DataFrame assumed to have at least one metric column")
+        if models_metrics[Columns.Model].isnull().values.any():
+            raise ValueError("Found NaN values in `Model`column of `metrics_data`")
```
feldlime, line 164:
```suggestion
            raise ValueError("Found NaN values in `Model` column of `metrics_data`")
```

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,274 @@ def construct(
...
+        if Columns.Split in models_metrics.columns and models_metrics[Columns.Split].isnull().values.any():
+            raise ValueError("Found NaN values in `Split`column of `metrics_data`")
```
feldlime, line 166:
```suggestion
            raise ValueError("Found NaN values in `Split` column of `metrics_data`")
```

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,274 @@ def construct(
...
+        ref_group_name = next(iter(grouped.groups.keys()))
+        ref_group = grouped.get_group(ref_group_name)
+        ref_fold_count = ref_group[Columns.Split].nunique()
+        ref_fold_names = set(ref_group[Columns.Split].unique())
+
+        for model, group in grouped:
+            if group["i_split"].nunique() != ref_fold_count:
+                raise ValueError(f"""{model} does not have the same fold amount as {ref_group_name}""")
+            if set(group["i_split"].unique()) != ref_fold_names:
+                raise ValueError(f"{model} does not have the same fold names as {ref_group_name}")
```
feldlime, line 184: Now errors look a bit weird: if some model has 2 splits and other 3 splits, why should we say that the 2nd has problems but not the 1st one?  
Let's just say that smth is wrong. And make the code simpler :)
```suggestion
        splits = df.groupby(Columns.Model)["split"].apply(lfrozenset)
        splits_set = set(splits)
        if splits_set > 1:
            raise ValueError(f"All models must have the same splits. But now they are different: {splits_set}")
```

feldlime: Also: we are taking unique splits here, but they should be unique anyway right? I mean there must not we 2 rows with the same model and split  
Let's check this as well if it's correct point

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,274 @@ def construct(
...
+
+    @staticmethod
+    def _validate_models_metrics_names(models_metrics: pd.DataFrame) -> None:
+        # Validate that all Models names are unique
+        models_metrics = models_metrics.copy()
+        if Columns.Split in models_metrics.columns:
+            models_names_comb = models_metrics[Columns.Model] + models_metrics[Columns.Split].astype(str)
+            if models_names_comb.nunique() != len(models_names_comb):
+                raise ValueError("Each `Model` value in the `metrics_data` DataFrame must be unique")
```
feldlime, line 192: It seems here should be different text of error, now it's the same with the `else` branch

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,274 @@ def construct(
...
+                raise ValueError(f"""{model} does not have the same fold amount as {ref_group_name}""")
+            if set(group["i_split"].unique()) != ref_fold_names:
+                raise ValueError(f"{model} does not have the same fold names as {ref_group_name}")
+
+    @staticmethod
+    def _validate_models_metrics_names(models_metrics: pd.DataFrame) -> None:
+        # Validate that all Models names are unique
+        models_metrics = models_metrics.copy()
+        if Columns.Split in models_metrics.columns:
+            models_names_comb = models_metrics[Columns.Model] + models_metrics[Columns.Split].astype(str)
```
feldlime, line 190: Really dangerous approach. Imagine you have next rows:  

Model | Split
------+------
a         | bc
------+-----
ab      | c

This approach won't work correctly. 
Better use `duplicated` or `groupby`

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,275 @@ def construct(
...
+
+    @staticmethod
+    def _validate_models_metrics_names(models_metrics: pd.DataFrame) -> None:
+        # Validate that all Models names are unique
+        if Columns.Split in models_metrics.columns:
+            models_names_comb = models_metrics[Columns.Model] + models_metrics[Columns.Split].astype(str)
+            if models_names_comb.nunique() != len(models_names_comb):
+                raise ValueError("Each `Model` value in the `metrics_data` DataFrame must be unique")
+        else:
+            if models_metrics[Columns.Model].nunique() != len(models_metrics[Columns.Model]):
```
feldlime, line 194:
```suggestion
            if models_metrics[Columns.Model].nunique() != len(models_metrics):
```

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,274 @@ def construct(
...
+        cls._validate_models_metadata(models_metadata)
 
-    @property
-    @lru_cache
-    def metric_names(self) -> tp.List[str]:
-        """List of metric column names from the `models_metrics`."""
-        non_metric_columns = {Columns.Split, Columns.Model}
-        return [col for col in self.models_metrics.columns if col not in non_metric_columns]
+        merged_data = models_metrics.merge(models_metadata, on=Columns.Model, how="left")
+        merged_data = merged_data.replace(",", ";", regex=True)
```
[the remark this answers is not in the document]
Gooogr: I replaced "," by ";" because I trim metadata in chart legend by ", " substring. So, I want to be sure that I wouldn't trim something incorrectly.

rectools/visuals/metrics_app.py
Gooogr: Yes, we discuss this type of check here: https://github.com/MobileTeleSystems/RecTools/pull/154#discussion_r1687973973

Would be refactored according to your approach

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,274 @@ def construct(
...
-        if len(self.models_metrics.columns) < 3:
+        if Columns.Split in self.data.columns:
+            return sorted(self.data[Columns.Split].unique())
+        return None
+
+    @staticmethod
+    def _validate_models_metrics_base(models_metrics: pd.DataFrame) -> None:
+        if Columns.Model not in models_metrics.columns:
+            raise KeyError("Missing `Model` column in `metrics_data` DataFrame")
+        if len([item for item in models_metrics.columns if item not in {Columns.Model, Columns.Split}]) < 1:
```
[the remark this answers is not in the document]
feldlime: @Gooogr Please do not mark any comments as `resolved` yourself, instead just write "done" or smth like this. We have inner rule that only the author can resolve the comment

rectools/visuals/metrics_app.py
Gooogr: Fixed

rectools/visuals/metrics_app.py
Gooogr: Fixed

rectools/visuals/metrics_app.py
Gooogr: Fixed

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,267 @@ def construct(
...
+            meta_value, model_name = splitted_row
+            return meta_value, model_name
+        return "", raw_string
+
+    def _create_chart_figure(
+        self,
+        data: pd.DataFrame,
+        metric_x: str,
+        metric_y: str,
+        color: str,
```
feldlime, line 222: Let's rename it to `color_col` please, now it's really confusing 

Same for metric_x and metric_y

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,267 @@ def construct(
...
         use_avg: widgets.Checkbox,
         fold_i: widgets.Dropdown,
-    ) -> None:
-        data = self._make_chart_data_avg() if use_avg.value else self._make_chart_data(fold_i.value)
-        existing_traces = {trace.name: trace for trace in fig_widget.data}
+        meta_feature: widgets.Dropdown,
+        use_meta: widgets.Checkbox,
+    ) -> None:  # pragma: no cover
+        chart_data = self._create_chart_data(use_avg, fold_i)
+        color_clmn = meta_feature.value if use_meta.value else Columns.Model
```
feldlime, line 257: Please use `col` or `column`, it's kinda stylecode :)

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New params (`tol`, `maxiter`, `random_state`) to the `PureSVD` model ([#130](https://github.com/MobileTeleSystems/RecTools/pull/130))
 - Recommendations data quality metrics: `SufficientReco`, `UnrepeatedReco`, `CoveredUsers` ([#155](https://github.com/MobileTeleSystems/RecTools/pull/155))
 - `r_precision` parameter to `Precision` metric ([#155](https://github.com/MobileTeleSystems/RecTools/pull/155))
+- Add metadata support for `MetricApp` ([#154](https://github.com/MobileTeleSystems/RecTools/pull/154))
```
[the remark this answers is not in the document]
Gooogr: Done

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,267 @@ def construct(
...
+
+        # Save dots symbols from the previous widget state
+        # Remove metainfo from trace name. Thus we guarantee to map with traces from previous state
+        model_name2symbol = {
```
feldlime, line 261: We don't use `smth2smth` namings usually in rectools, and especially don't mix 2 different naming styles in one variable name. Please rename

Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
Gooogr: Done

rectools/visuals/metrics_app.py
```diff
@@ -36,28 +38,32 @@ class MetricsApp:
...
+        data: pd.DataFrame,
+        metric_names: tp.List[str],
+        meta_names: tp.List[str],
         show_legend: bool = True,
         auto_display: bool = True,
-        layout_kwargs: tp.Optional[tp.Dict[str, tp.Any]] = None,
+        scatter_kwargs: tp.Optional[tp.Dict[str, tp.Any]] = None,
     ):
-        self.models_metrics = models_metrics
+        self.data = data
```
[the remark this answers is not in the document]
Gooogr: Done

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
feldlime: Ok, then please let's do next
1. Comment it here
2. Choose more rare symbol, say `|`
3. Put it into a constant like `META_MODEL_SEP = ' || '` (It's intentionally surrounded with spaces, you can remove them if you want, but my point is to use the same string everywhere. Now you're replacing `,`, but your actual separator is `, `
4. Add another constant like `META_MODEL_SEP_REPLACEMENT = ' |; '` for replacement
5. Use this constants everywhere: when replacing, when joining, when splitting
6. In `_split...` function please rename `splitter` to `sep`

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,267 @@ def construct(
...
+        return self.data[self.data[Columns.Split] == fold_number].reset_index(drop=True)
 
     @lru_cache
     def _make_chart_data_avg(self) -> pd.DataFrame:
-        return self.models_metrics.drop(columns=Columns.Split).groupby(Columns.Model, sort=False).mean().reset_index()
+        metric_data_columns = [Columns.Model] + self.metric_names
+        meta_data_columns = [Columns.Model] + self.meta_names
+        # separate metric data because meta data could contain strings
+        metrics_data = self.data[metric_data_columns].groupby([Columns.Model], sort=False).mean().reset_index()
+        meta_data = self.data[meta_data_columns].drop_duplicates()
```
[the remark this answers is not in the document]
feldlime: Understood, thanks.

Btw, what was the point to merge them if you need to separate after it?

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,264 @@ def construct(
...
+        self, use_avg: widgets.Checkbox, fold_i: widgets.Dropdown
+    ) -> pd.DataFrame:  # pragma: no cover
+        if use_avg.value or fold_i.value is None:
+            return self._make_chart_data_avg()
+        return self._make_chart_data_fold(fold_i.value)
+
     def display(self) -> None:
         """Display MetricsApp widget"""
         metric_x = widgets.Dropdown(description="Metric X:", value=self.metric_names[0], options=self.metric_names)
         metric_y = widgets.Dropdown(description="Metric Y:", value=self.metric_names[-1], options=self.metric_names)
```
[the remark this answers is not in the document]
Gooogr: The thing is that this widget can work even with one metric - we can visualize the distance between points along the selected metric axis.

I think this case is exotic, but not useless.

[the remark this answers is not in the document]
feldlime: It may fail with IndexError, but there is no sense to build this plot if there is only 1 metric, right? We can add a check that we have at least 2 metrics.

Or just take smth like `self.metric_names[min(1, len(metric_names) - 1)]`.

My point was it's a bit weird to take the last value by default

Gooogr:
@feldlime 
Sorry, but for some reason I can't write comments under https://github.com/MobileTeleSystems/RecTools/pull/154#discussion_r1688412110  topic.
Correct me if I'm wrong, but if we do the replacement as you suggest, it turns out that if the user adds a model named “my model, some text, k=10” to the dataframe, it will be “my model ||  some text ||  k=10” in the legend on the chart. Or it's OK?
That's why I used the ";" symbol as a more neutral replacement.

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
Gooogr: Yes, I see your point. Updated tests, now we need at least two metric columns. Set index as [0] and [1] accordingly.

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
Gooogr: Refactor function to avoid all these drops and merging. 
Now we aggregate all columns in one step.

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
feldlime: No, why  
According to my suggestion `“my model, some text, k=10”` will be kept as is and in the legend you will have `some meta || my model, some text, k=10`.  
But if user has model named `"my model || some text || k=10"`, then in legend they will have `some meta || my model |; some text |; k=10` which is not very good looking but okaish (you can use smth better than `|;`)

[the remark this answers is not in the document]
feldlime: I suggest not to touch and not to use commas at all since they are quite common

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
feldlime: I just noticed that I wrote my comment after yours, didn't see it before. But nevertheless, it's still relevant.  
Let's handle case with 1 metric somehow, but let's take first 2 metrics by default, not first and last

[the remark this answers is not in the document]
Gooogr: Ok, set minimal amount of metric columns back to one and handle index issue as you suggested above.

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
Gooogr: Done

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +97,264 @@ def construct(
...
-        use_avg = widgets.Checkbox(description="Avg folds", value=True)
-        fold_i = widgets.Dropdown(description="Fold number:", value=self.fold_ids[0], options=self.fold_ids)
+        use_avg = widgets.Checkbox(description="Average folds", value=True)
+        fold_i = widgets.Dropdown(
+            description="Fold number:",
+            value=self.fold_ids[0] if self.fold_ids is not None else -1,
```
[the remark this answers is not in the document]
Gooogr: 1. Unfortunately widget would through error in this case. That's why I use -1 for initialization. But user will not see this.
2. We can and we did it.

At this step we initialize all widgets, but then hide some of them before rendering.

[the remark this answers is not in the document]
feldlime: 1. But you're using `None` for the `meta_feature` dropdown. Or is it different in some way?
2. Cool!

rectools/visuals/metrics_app.py
```diff
@@ -25,7 +26,10 @@
 WIDGET_WIDTH = 800
 WIDGET_HEIGHT = 500
 TOP_CHART_MARGIN = 20
-LEGEND_TITLE = "Models"
+DEFAULT_LEGEND_TITLE = "model name"
+NAN_COLOR = "grey"
+META_MODEL_SEP = "❘"
```
feldlime, line 31: Do you think it makes sense to add spaces around for readability? (If yes, then also for the next one)

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
feldlime: Not sure I noticed the change :(

What I see:
1. We take 2 tables in construct. `meta_data` has unique rows
2. We merge them
3. Here we split them. `meta_data` has duplicates now
4. We aggregate metrics
5. We remove duplicates from `meta_data`
6. We merge them again

Looks super weird to me.  
What's the reason? Why don't we want to keep 2 separate tables?
It'd be conceptually more correct, and 2 times simpler in this case:
1. We take 2 tables in construct. `meta_data` has unique rows
2. We aggregate metrics
3. We merge them

rectools/visuals/metrics_app.py
```diff
@@ -85,125 +104,275 @@ def construct(
...
+            x=x_col,
+            y=y_col,
+            color=color_col,
+            symbol=Columns.Model,
+            **scatter_kwargs,
+        )
+        # Set custom legend for meta info coloring
+        if color_col != Columns.Model:
+            for trace, meta_value, model_name in zip(fig.data, data[color_col], data[Columns.Model]):
+                trace.name = f"{meta_value}{META_MODEL_SEP}{model_name}"
```
feldlime, line 253: Just out of curiosity: didn't `labels` argument help?

feldlime: And let's maybe swap model name and meta value

blondered: Please don't swap. Meta value will not be readable if we swap

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
Gooogr: Sorry, seems like that commit was lost after force push. I updated` _make_chart_data_avg` again.
We discussed pros and cons of merging columns data and meta data tables here: https://github.com/MobileTeleSystems/RecTools/pull/154#discussion_r1649685194

We can discuss it again, if we need. 
Looking at the code now, I'd say that an additional metadata merge on a table with metrics averages (or a slice by fold number) would look more readable.

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
Gooogr: You right, I check it again and we could use `None` and `[]` for fold_id. Code was updated.

rectools/visuals/metrics_app.py
Gooogr: I checked both options, and personally prefer version without additional spaces. Let me show you:
Without additional spaces:
Screenshot from 2024-07-26 10-53-05
With them:
Screenshot from 2024-07-26 10-52-30

rectools/visuals/metrics_app.py
feldlime: discussed, agreed not to swap

rectools/visuals/metrics_app.py
feldlime: Ok, doesn't  matter choose whatever you prefer, both looks fine for me

rectools/visuals/metrics_app.py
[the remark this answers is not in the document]
feldlime: Now looks much better, thanks!

At that moment I didn't know how it would turn out. Let's keep it, enough discussions 😅

blondered (approved):

PR #177, opened 2024-08-10
Removed internal ids usage outside models

blondered:
- Removed `assume_external_ids` parameter in `recommend` and `recommend_to_items` model methods 
- Added `on_unsupported_targets` in `recommend` and `recommend_to_items` model methods 
- Added `filter_on_interactions_df_row_indexes` method of `Dataset`
- Fixed `IntraListDiversity` metric computation in `cross_validate`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `Debias` mechanism for classification, ranking and auc metrics. New parameter `is_debiased` to `calc_from_confusion_df`, `calc_per_user_from_confusion_df` methods of classification metrics, `calc_from_fitted`, `calc_per_user_from_fitted` methods of auc and rankning (`MAP`) metrics, `calc_from_merged`, `calc_per_user_from_merged` methods of ranking (`NDCG`, `MRR`) metrics. ([#152](https://github.com/MobileTeleSystems/RecTools/pull/152))
 - `filter_on_interactions_df_row_indexes` method of `Dataset` ([#177](https://github.com/MobileTeleSystems/RecTools/pull/177))
+- `on_unsupported_targets` parameter to  `recommend` and `recommend_to_items` model methods ([#177](https://github.com/MobileTeleSystems/RecTools/pull/177))
 
 ### Removed
 - [Breaking] `assume_external_ids` parameter in `recommend` and `recommend_to_items` model methods ([#177](https://github.com/MobileTeleSystems/RecTools/pull/177))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `Debias` mechanism for classification, ranking and auc metrics. New parameter `is_debiased` to `calc_from_confusion_df`, `calc_per_user_from_confusion_df` methods of classification metrics, `calc_from_fitted`, `calc_per_user_from_fitted` methods of auc and rankning (`MAP`) metrics, `calc_from_merged`, `calc_per_user_from_merged` methods of ranking (`NDCG`, `MRR`) metrics. ([#152](https://github.com/MobileTeleSystems/RecTools/pull/152))
 - `nbformat >= 4.2.0` dependency to `[visuals]` extra ([#169](https://github.com/MobileTeleSystems/RecTools/pull/169))
-- `filter_on_interactions_df_row_indexes` method of `Dataset` ([#177](https://github.com/MobileTeleSystems/RecTools/pull/177))
+- `filter_interactions` method of `Dataset` ([#177](https://github.com/MobileTeleSystems/RecTools/pull/177))
 - `on_unsupported_targets` parameter to  `recommend` and `recommend_to_items` model methods ([#177](https://github.com/MobileTeleSystems/RecTools/pull/177))
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `Debias` mechanism for classification, ranking and auc metrics. New parameter `is_debiased` to `calc_from_confusion_df`, `calc_per_user_from_confusion_df` methods of classification metrics, `calc_from_fitted`, `calc_per_user_from_fitted` methods of auc and rankning (`MAP`) metrics, `calc_from_merged`, `calc_per_user_from_merged` methods of ranking (`NDCG`, `MRR`) metrics. ([#152](https://github.com/MobileTeleSystems/RecTools/pull/152))
+- `filter_on_interactions_df_row_indexes` method of `Dataset`
+
+### Removed
+- [Breaking] `assume_external_ids` parameter in `recommend` and `recommend_to_items` model methods 
+
+### Fixed
+- `IntraListDiversity` metric computation in `cross_validate`
 
 ## [0.7.0] - 29.07.2024
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `Debias` mechanism for classification, ranking and auc metrics. New parameter `is_debiased` to `calc_from_confusion_df`, `calc_per_user_from_confusion_df` methods of classification metrics, `calc_from_fitted`, `calc_per_user_from_fitted` methods of auc and rankning (`MAP`) metrics, `calc_from_merged`, `calc_per_user_from_merged` methods of ranking (`NDCG`, `MRR`) metrics. ([#152](https://github.com/MobileTeleSystems/RecTools/pull/152))
-- `filter_on_interactions_df_row_indexes` method of `Dataset`
+- `filter_on_interactions_df_row_indexes` method of `Dataset` ([#177](https://github.com/MobileTeleSystems/RecTools/pull/177))
 
 ### Removed
-- [Breaking] `assume_external_ids` parameter in `recommend` and `recommend_to_items` model methods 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -11,9 +11,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Fixed
 - `display()` method in `MetricsApp` ([#169](https://github.com/MobileTeleSystems/RecTools/pull/169))
+- `IntraListDiversity` metric computation in `cross_validate` ([#177](https://github.com/MobileTeleSystems/RecTools/pull/177))
+
+### Removed
+- [Breaking] `assume_external_ids` parameter in `recommend` and `recommend_to_items` model methods ([#177](https://github.com/MobileTeleSystems/RecTools/pull/177))
+
+
+### Fixed
```
feldlime, line 25: Duplicated 2 lines

rectools/dataset/dataset.py
```diff
@@ -1,4 +1,4 @@
-#  Copyright 2022-2024 MTS (Mobile Telesystems)
+#  Copyright 2022 MTS (Mobile Telesystems)
```
feldlime, line 1: Why? Did you use the script?

rectools/dataset/dataset.py
```diff
@@ -245,3 +246,67 @@ def get_raw_interactions(self, include_weight: bool = True, include_datetime: bo
         pd.DataFrame
         """
         return self.interactions.to_external(self.user_id_map, self.item_id_map, include_weight, include_datetime)
+
+    def filter_on_interactions_df_row_indexes(
+        self,
+        row_indexes_to_keep: np.ndarray,
```
feldlime, line 252: Since we refused python 3.7 support, let's start adding types for numpy arrays. Please specify here that this is array of integers

rectools/model_selection/cross_validate.py
```diff
@@ -1,71 +1,15 @@
-#  Copyright 2023-2024 MTS (Mobile Telesystems)
```
feldlime, line 1: Why did you remove the license header?

rectools/model_selection/cross_validate.py
```diff
@@ -132,34 +86,23 @@ def cross_validate(  # pylint: disable=too-many-locals
...
-        test_users = interactions_df_test[Columns.User].unique()  # 1x internal
-        catalog = interactions_df_train[Columns.Item].unique()  # 1x internal
-
-        if items_to_recommend is not None:
-            item_ids_to_recommend = dataset.item_id_map.convert_to_internal(
-                items_to_recommend, strict=False
-            )  # 1x internal
-        else:
-            item_ids_to_recommend = None
+        interactions_df_test = dataset.get_raw_interactions().iloc[test_ids]
```
feldlime, line 102: Seems really ineffective:
1. You can get raw interactions once before the cycle.
2. We invented internal ids to avoid memory errors caused by big external ids, and here this logic fails.

What can be done now: you can use your new method `filter_on_interactions_df_row_indexes` to get test dataset, and then extract raw interactions from it. It makes sense since test part is usually much smaller than dataset. Yo can also add an argument to your method like `keep_features` and set it to `False` here to make it more effective.

What can we do in future: avoid converting anything to external ids and calculate metrics on internal ones.

rectools/models/base.py
```diff
@@ -1,4 +1,4 @@
-#  Copyright 2022-2024 MTS (Mobile Telesystems)
+#  Copyright 2022 MTS (Mobile Telesystems)
```
feldlime, line 1: ?

tests/models/test_base.py
```diff
@@ -461,16 +342,69 @@ def test_full_model_works_for_all_without_features(self, kind: str) -> None:
...
-        targets = self.colds[kind]
+    def test_not_cold_models_with_cold_targets(self, dataset_key: str, kind: str, model_key: str) -> None:
+        targets = self.colds[kind] + self.hots[kind]
+
+        # raise
         with pytest.raises(ValueError, match="doesn't support recommendations for cold"):
-            self._get_reco(targets, model_key, dataset_key, kind)
+            self._get_reco(targets, model_key, dataset_key, kind, on_unsupported_targets="raise")
+
+        # ignore
```
feldlime, line 352: Please NEVER test multiple cases in one test.  Ask me if you want me to explain why.    
Please, either write multiple tests, or use `parametrize` or `subtest`

CHANGELOG.md
blondered: Fixed

rectools/dataset/dataset.py
blondered: Sorry. Fixed

rectools/dataset/dataset.py
blondered: Decided to do it later

rectools/model_selection/cross_validate.py
blondered: Fixed

rectools/model_selection/cross_validate.py
blondered: I did the first one.
Can't see any optimization with calling `filter_on_interactions_df_row_indexes` for test interactions. It calls `get_raw_interactions` under the hood

rectools/models/base.py
blondered: Fixed

tests/models/test_base.py
blondered: Fixed

rectools/model_selection/cross_validate.py
feldlime: First one was the obvious thought but it still not enough.  We can't keep raw interactions in memory.

feldlime: > Can't see any optimization with calling filter_on_interactions_df_row_indexes for test interactions. It calls get_raw_interactions under the hood

Conclusion: we need to rewrite `filter_on_interactions_df_row_indexes`

blondered: What do you want to do? get internal interactions.df, apply [test_ids] and then convert to external?

blondered: Fixed

rectools/dataset/dataset.py
```diff
@@ -245,3 +246,66 @@ def get_raw_interactions(self, include_weight: bool = True, include_datetime: bo
...
+            Otherwise internal -> 2x internal ids mapping will be created.
+        keep_features_for_removed_entities : bool, default `True`
+            Whether to keep all features for users and items that are not hot any more.
+
+        Returns
+        -------
+        Dataset
+            Filtered dataset that has only selected interactions, new ids mapping and processed features.
+        """
+        interactions_df = self.interactions.df.loc[row_indexes_to_keep]
```
feldlime, line 276: Why did you change `iloc` to `loc`? Sure it's safe? We don't pay attention to indices usually

feldlime: Btw you wrote in docstring `Original dataset interactions df row indexes` but we don't keep original indexes, so they are row numbers actually

rectools/model_selection/cross_validate.py
```diff
@@ -132,34 +100,26 @@ def cross_validate(  # pylint: disable=too-many-locals
...
-        # We need to avoid fitting models on sparse matrices with all zero rows/columns =>
-        # => we need to create a fold dataset which contains only hot users and items for current training
-        fold_dataset = _gen_2x_internal_ids_dataset(
-            interactions_df_train, dataset.user_features, dataset.item_features, prefer_warm_inference_over_cold
+        fold_dataset = dataset.filter_on_interactions_df_row_indexes(
+            row_indexes_to_keep=train_ids,
+            keep_external_ids=True,
+            keep_features_for_removed_entities=prefer_warm_inference_over_cold,
         )
+        interactions_df_test = dataset.interactions.df.loc[test_ids]
```
feldlime, line 116: Why don't you use your new dataset method here?

rectools/dataset/dataset.py
blondered: `filter_on_interactions_df_row_numbers`?

Changed to iloc back

feldlime: I'd name it shorter since it's a public method. E.g. `iloc_interactions`

rectools/model_selection/cross_validate.py
feldlime: But it'd be just one line of code instead of 3, no?

blondered: It would be 2 lines here instead of 3.
5 lines in the func itself  and 20 lines for tests :)
right now the logic is clear and all is fine. no need for dataset when we need  just interactions

rectools/dataset/dataset.py
blondered: That's not good since we'll get a dataset from this method (not interactions). And we will write more filters for dataset in the future (like filter on features selected)
so `filter_on` beginning is very reasonable

feldlime: ok, then let's just name it `filter` (I don't really understand `on`)

rectools/model_selection/cross_validate.py
feldlime: Didn't get your point, sorry. 
You have the func already, just call it and take interactions.  
It'd be more readable than doing manual manipulations.  
Why do you need add lines to the func?

blondered: It's not appropriate since new func is for dataset creation. It has more going on than just converting ids of interactions to external ones. For example it will construct new ids maps. We just don't need it here

rectools/dataset/dataset.py
blondered: ok)
renamed to `filter_interactions_df_rows`

feldlime: You said we're going to filter features with this method in the future - let's then call it just `filter`?

blondered: I thought there will be different `filter_<on something>` methods. 
I'm not sure if we want to do all filters in the same method. It's probably a good idea - this way we never need to call a chain of filters (which is inefficient). So you're probably right.

It's gonna be a `filter` method. And possible arguments might be (just my imagination): `interactions_df_rows`, `user_features_subset`, `item_features_subset` and so on (if any). Right?

feldlime: yes, sounds reasonable

blondered: Fixed

feldlime (approved):

PR #178, opened 2024-08-13
Configs for simple models

mikesokolovv:
## Description

- Added model configs for `EASE`, `Random` and `PureSVD` models
- Moved `RandomState` serialization to models/base.py

## Type of change

- [ ] Optimization

### DIFF
--- a/rectools/models/ease.py
+++ b/rectools/models/ease.py
@@ -68,18 +68,14 @@ class EASEModel(ModelBase[EASEModelConfig]):
         num_threads: int = 1,
         verbose: int = 0,
     ):
-        self._config = self._make_config(regularization, num_threads, verbose)
 
         super().__init__(verbose=verbose)
         self.weight: np.ndarray
         self.regularization = regularization
         self.num_threads = num_threads
 
-    def _make_config(self, regularization: float, num_threads: int, verbose: int) -> EASEModelConfig:
-        return EASEModelConfig(regularization=regularization, num_threads=num_threads, verbose=verbose)
-
     def _get_config(self) -> EASEModelConfig:
-        return self._config
+        return EASEModelConfig(regularization=self.regularization, num_threads=self.num_threads, verbose=self.verbose)
 
     @classmethod
     def _from_config(cls, config: EASEModelConfig) -> tpe.Self:
--- a/rectools/models/ease.py
+++ b/rectools/models/ease.py
@@ -17,17 +17,26 @@
 import typing as tp
 
 import numpy as np
+import typing_extensions as tpe
 from scipy import sparse
 
 from rectools import InternalIds
 from rectools.dataset import Dataset
+from rectools.models.base import ModelConfig
 from rectools.types import InternalIdsArray
 
 from .base import ModelBase, Scores
 from .rank import Distance, ImplicitRanker
 
 
-class EASEModel(ModelBase):
+class EASEModelConfig(ModelConfig):
+    """Config for `EASE` model."""
+
+    regularization: float = 500.0
+    num_threads: int = 1
+
+
+class EASEModel(ModelBase[EASEModelConfig]):
     """
     Embarrassingly Shallow Autoencoders for Sparse Data model.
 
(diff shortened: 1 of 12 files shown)


feldlime:
Hi, thanks for PR!  
Could you please give it a more relevant name? "Additional" is quite ambiguous; for me, it sounds like "optional."  
Also, please separate changes in the description into 2 bullet points (added configs and moved rs)

rectools/models/ease.py
```diff
@@ -51,17 +60,31 @@ class EASEModel(ModelBase):
...
 
+    config_class = EASEModelConfig
+
     def __init__(
         self,
         regularization: float = 500.0,
         num_threads: int = 1,
         verbose: int = 0,
     ):
+        self._config = self._make_config(regularization, num_threads, verbose)
```
feldlime, line 71: It seems we don't really need it since we keep both parameters as attributes. Take a look of how it's done for knn model

rectools/models/pure_svd.py
```diff
@@ -59,6 +72,7 @@ def __init__(
         random_state: tp.Optional[int] = None,
         verbose: int = 0,
     ):
+        self._config = self._make_config(factors, tol, maxiter, random_state, verbose)
```
feldlime, line 75: Same here

rectools/models/random.py
```diff
@@ -70,13 +78,26 @@ class RandomModel(ModelBase):
     recommends_for_warm = False
     recommends_for_cold = True
 
+    config_class = RandomModelConfig
+
     def __init__(self, random_state: tp.Optional[int] = None, verbose: int = 0):
+        self._config = self._make_config(random_state, verbose)
```
feldlime, line 84: And here

tests/models/test_ease.py
```diff
@@ -27,6 +27,31 @@
 
 
 class TestEASEModel:
+    def test_from_config(self) -> None:
```
feldlime, line 30: Please make the tests in the same way as they are done for knn and also models:  
1. Separate class
2. Add `test_get_config_and_from_config_compatibility` and `test_default_config_and_default_model_params_are_the_same`. For these 2 tests I think it makes sense to add some common functions to utils, like `assert_second_fit_refits_model`

rectools/models/ease.py
mikesokolovv: _make_config method was removed.

rectools/models/pure_svd.py
mikesokolovv: Done.

tests/models/test_ease.py
mikesokolovv: Separate class was created for each model as per provided example. 
2 functions were added to `tests/utils.py` to cover compatibility and similarity tests.

feldlime: Great! Could you please use your new functions in ALS and KNN model tests?

feldlime:
> Could you please give it a more relevant name? "Additional" is quite ambiguous; for me, it sounds like "optional." 

By this I wanted to say that "additional" sounds like "optional", but these configs are not "optional" and hence not "additional". So, please, give to this PR some meaningful name, and better do not connect it with the branch name since it's already different. You can name it `Configs for simple models`, it'd be fine.  
Sorry for confusion.

tests/models/utils.py
```diff
@@ -47,3 +48,28 @@ def assert_second_fit_refits_model(
...
+
+def assert_default_config_and_default_model_params_are_the_same(
+    model: ModelBase, default_config: tp.Dict[str, tp.Any]
+) -> None:
+    model_from_config = model.from_config(default_config)
+    assert model_from_config.get_config() == model.get_config()
+
+
+def assert_get_config_and_from_config_compatibility(
+    model: ModelBase, dataset: Dataset, initial_config: tp.Dict[str, tp.Any], simple_types: tp.Optional[bool] = None
```
feldlime, line 61: 1. No sense to make it `simple_types: tp.Optional[bool] = None`, it's better to make it `simple_types: bool = False` as it is by default in `get_config`, then you don't need `if simple_types is not None ...`
2. It's better to check both cases for all models (including your new models), so you never need `None`

mikesokolovv: Fair point, added simple_types support into each model test.

feldlime (approved):

PR #179, opened 2024-08-13
Implement `fit_partial()` for `ImplicitALSWrapperModel` and `LightFMWrapperModel`

### LINKED ISSUE #176: fit_partial function for ImplicitALSWrapperModel and LightFMWrapperModel
### Feature Description

Implicit's ALS and LightFM provide partial_fit capability. 
https://benfred.github.io/implicit/api/models/cpu/als.html#implicit.cpu.als.AlternatingLeastSquares.partial_fit_users
https://benfred.github.io/implicit/api/models/cpu/als.html#implicit.cpu.als.AlternatingLeastSquares.partial_fit_items
https://making.lyst.com/lightfm/docs/lightfm.html#lightfm.LightFM.fit_partial

It would be great if RecTools had a standardized way to use them for incremental model updates.

### Why this feature?

If we can use partial_fit, we can avoid full re-training every time. This is beneficial when the original model is trained by a massive dataset.

### Additional context

_No response_

chezou:
## Description

Introduce fit_partial for ImplicitALSWrapperModel and LightFMWrapperModel.

Expected usage:

```py
dataset = ...

new_interactions = ...

new_dataset = dataset.rebuild_with_new_data(interactions_df=new_interactions)
model.fit_partial(new_dataset)
```

Due to the complexity of explicit user/item feature fitting of ALS, I don't support them for fit_partial to ALS with features for now.

Close #176 

## Type of change

- [ ] Optimization

## How Has This Been Tested?
  - Yes
  - Yes
  - Yes
  - Yes


rectools/models/implicit_als.py
```diff
@@ -90,6 +90,25 @@ def _fit(self, dataset: Dataset) -> None:  # type: ignore
                 self.verbose,
             )
 
+    def _fit_partial(self, dataset: Dataset) -> None:
+        # deepcopy does not copy model.item_factors and model.user_factors.
```
feldlime, line 94: If you call `fit_partial` now on just initialised model, you'll get an error on the last rows of this method since `self.model` is not defined. What is the expected behavior in this case? Let's either support calling `fit_partial` without calling `fit` before or raise a reasonable error

tests/models/test_implicit_als.py
```diff
@@ -346,3 +346,63 @@ def test_i2i_with_warm_and_cold_items(self, use_gpu: bool, dataset: Dataset) ->
...
+            [150, 12],
+            [150, 15],
+        ]
+        new_interactions = pd.DataFrame(data, columns=Columns.UserItem)
+        new_interactions[Columns.Weight] = 1
+        new_interactions[Columns.Datetime] = "2021-09-10"
+        new_dataset = dataset.construct_new_datasets(new_interactions)
+        model.fit_partial(new_dataset)
+        actual = model.recommend(
+            users=[150],  # new user
```
feldlime, line 364: Could you please request here recomendation for one user from original dataset as well? To ensure we didn't lose its data (same for lightfm)

feldlime (commented):
Thanks a lot for such a great feature! 

I added some comments, if you don't agree with them, let's discuss. I'm not 100% sure I understand the use case, so may be some of the comments are wrong. But if you agree with them, then it's fine.

rectools/dataset/dataset.py
```diff
@@ -245,3 +258,72 @@ def get_raw_interactions(self, include_weight: bool = True, include_datetime: bo
         pd.DataFrame
         """
         return self.interactions.to_external(self.user_id_map, self.item_id_map, include_weight, include_datetime)
+
+    def construct_new_datasets(
```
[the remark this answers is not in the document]
blondered: I'd call it `build_new_dataset_on_id_map` or smth like it. But we need to think about the conception first.

Right now it's not working properly in the conceptions that we use. We have a rule that all hot users (users that have interactions) always go before warm users (users that have only features). And we use this assumption in some of our impertant internal logic.

New method can break this rule because new users with interactions will have ids after warm users from previous dataset state. So we really can't make this a public method in the library.

Options are:
- Rewrite logic. We need to keep only hot users ids from previous state. Then add new hot users ids. Then add new warm user ids. I suppose we are dropping all previous warm users because we are totally changing the actual data. (All the same is for items) In this case I'd call it `update_with_new_data` or `rebuild_with_new_data` or `construct_with_new_data`
- Just don't make it a public method. Move it to the tests if we need it there. But then again write a correct public method for required logic. Also tests will be needed for it anyway

rectools/dataset/dataset.py
```diff
@@ -91,6 +92,14 @@ def get_hot_item_features(self) -> tp.Optional[Features]:
             return None
         return self.item_features.take(range(self.n_hot_items))
 
+    def get_hot_users(self) -> InternalIdsArray:
+        """Return internal ids of hot users."""
+        return self.interactions.df[Columns.User].unique()
```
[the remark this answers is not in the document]
chezou: The intention of recalculating here is to list hot users in the new interactions. Implicit and LightFM assume input weights for new users.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - `Debias` mechanism for classification, ranking and auc metrics. New parameter `is_debiased` to `calc_from_confusion_df`, `calc_per_user_from_confusion_df` methods of classification metrics, `calc_from_fitted`, `calc_per_user_from_fitted` methods of auc and rankning (`MAP`) metrics, `calc_from_merged`, `calc_per_user_from_merged` methods of ranking (`NDCG`, `MRR`) metrics. ([#152](https://github.com/MobileTeleSystems/RecTools/pull/152))
 - `nbformat >= 4.2.0` dependency to `[visuals]` extra ([#169](https://github.com/MobileTeleSystems/RecTools/pull/169))
+- Implement `fit_partial()` for `ImplicitALSWrapperModel` and `LightFMWrapperModel` ([#179](https://github.com/MobileTeleSystems/RecTools/pull/179))
```
blondered, line 14: I'm very excited about the new feature. Thank you for the PR!

Let's discuss what we're trying to achieve with `fit_partial`.
In implicit ALS has `partial_fit_users` and `partial_fit_items` methods that allow to update weights only for specific parts of users/items and also support feeding new users and items to already fitted model.
LightFM model has different logic. It has `fit_partial` method. From docs: "Unlike fit, repeated calls to this method will cause training to resume from the current model state". This method does not support adding new users/items. And it is not meant to update only parts of embeddings. 

What exactly do we want to achieve with `fit_partial` in rectools? I would say that first step would be to follow LightFM logic. We are missing the ability to resume training, out `fit` always refits models from the beginning, just like LightFM `fit`.

As for `partial_fit_users` and `partial_fit_items` - these are very model specific ALS methods. I wouldn't call them `fit_partial`. I would move them to a separate issue and PR.

@chezou @feldlime what do you think?

tests/models/test_implicit_als.py
chezou: Thanks, fixed ea33c05

rectools/models/implicit_als.py
chezou: Raising an error explicitly ea33c05

rectools/models/lightfm.py
```diff
@@ -89,6 +90,23 @@ def _fit(self, dataset: Dataset) -> None:  # type: ignore
             verbose=self.verbose > 0,
         )
 
+    def _fit_partial(self, dataset: Dataset, *args: Any, **kwargs: Any) -> None:  # type: ignore
+        self.model = deepcopy(self._model)
+
+        ui_coo = dataset.get_user_item_matrix(include_weights=True).tocoo(copy=False)
+        user_features = self._prepare_features(dataset.get_hot_user_features(), dataset.n_hot_users)
+        item_features = self._prepare_features(dataset.get_hot_item_features(), dataset.n_hot_items)
+
```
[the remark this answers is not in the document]
chezou: Thanks, fixed ea33c05

rectools/models/lightfm.py
```diff
@@ -89,6 +90,23 @@ def _fit(self, dataset: Dataset) -> None:  # type: ignore
...
+        self.model.fit_partial(
+            ui_coo,
+            user_features=user_features,
+            item_features=item_features,
+            sample_weight=ui_coo,
+            epochs=self.n_epochs,
```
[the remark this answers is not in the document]
chezou: Thanks, fixed ea33c05

rectools/models/lightfm.py
```diff
@@ -89,6 +90,23 @@ def _fit(self, dataset: Dataset) -> None:  # type: ignore
             verbose=self.verbose > 0,
         )
 
+    def _fit_partial(self, dataset: Dataset, *args: Any, **kwargs: Any) -> None:  # type: ignore
+        self.model = deepcopy(self._model)
```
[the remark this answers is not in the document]
chezou: Thanks, fixed ea33c05

CHANGELOG.md
chezou: Good point.

My motivation for adding `fit_partial` is to enable incremental training. For example, weekly batches for full training and daily incremental training to adopt new and existing user transactions.

I don't seriously care about preserving old users/items. Rather, shortening training time would be important.

Actually, LightFM has a known workaround for adding a new user/item. https://github.com/lyst/lightfm/issues/347#issuecomment-707829342 
I included that workaround into the patch. ea33c05

rectools/dataset/dataset.py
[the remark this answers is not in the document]
chezou: Thanks for pointing it out. I wasn't aware of the warm users' assumption. And yeah, I expected this to be a controversial change. And that's why I haven't written the test yet 😅 

Originally, I expected for RecTools' `Dataset` to do similar to LightFM's `Dataset.fit_partial()`, but I found that RecTools `Dataset` is frozen.

I'm okay to make it private, or I can rebuild `IdMap` by concatenating new and old users/items something like:

```py
old_hot_user_id_map = IdMap.from_dict({e: i for e,i in  zip(ds.user_id_map.convert_to_external(ds.get_hot_users()), ds.get_hot_users())})
old_hot_item_id_map = ...

new_user_id = old_hot_user_id_map.add_ids(...
```

I'm not confident if it's okay to drop the previous warm user yet though,.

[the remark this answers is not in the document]
chezou: Hmm, I attempt to drop old warm ids, but I face `UnknownIdError`.
https://github.com/MobileTeleSystems/RecTools/blob/f73d0541d6bef7291d149311f7be44dfb069ff7a/rectools/dataset/features.py#L121

Maybe, storing hot/warm marks like `hot_ids` in IdMap can be an option like `np.array([1, 0, 1])` where `1` represents hot and `0` represents warm id. Then, we can fetch hot or warm users/items explicitly.

CHANGELOG.md
chezou: Sounds good. Let me add tests for re-training with the same dataset.

rectools/dataset/dataset.py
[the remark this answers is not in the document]
chezou: Okay, let me revive my dropping code and see what it happens on CI. I'll share it later.

CHANGELOG.md
chezou: Added the tests 117c5c7

rectools/dataset/dataset.py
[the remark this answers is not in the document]
chezou: fixed to use `np.arange()` cd9ec52

rectools/dataset/dataset.py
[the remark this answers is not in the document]
chezou: @blondered I think I completed implementing the "Rewrite logic" option, but let me know if you find anything else.

CHANGELOG.md
chezou: @blondered I think https://github.com/MobileTeleSystems/RecTools/pull/179/commits/117c5c7a1a177d7820b530843a16a6635df9ed87 shows how it works for the use case 2.

Please let me know if you'd like to add anything else.

chezou:
@blondered @feldlime I think all comments are addressed. Please have a look and let me know if you need anything.

rectools/models/implicit_als.py
```diff
@@ -90,6 +90,26 @@ def _fit(self, dataset: Dataset) -> None:  # type: ignore
                 self.verbose,
             )
 
+    def _fit_partial(self, dataset: Dataset) -> None:
```
blondered, line 93: We've discussed it a lot. Let's not do ALS support for `fit_partial` for now. 
Main reasons are:
1. ALS wrapper has different behaviour then LightFM wrapper in current implementation. ALS cannot do `fit_partial` on model that has not yet been fitted. We can't leave it as it is. 
2. Features support should be here for ALS. So we're gonna have to rewrite some of our code for features support in `fit` to add support for `fit_partial` and also for `epochs` parameter. And it's a long story. It should be in another PR.

tests/models/test_lightfm.py
```diff
@@ -222,6 +222,140 @@ def test_with_weights(self, interactions_df: pd.DataFrame) -> None:
...
+                Columns.Rank: [1, 2, 1, 2, 1, 2],
+            }
+        )
+        pd.testing.assert_frame_equal(actual.drop(columns=Columns.Score), expected)
+        pd.testing.assert_frame_equal(
+            actual.sort_values([Columns.User, Columns.Score], ascending=[True, False]).reset_index(drop=True),
+            actual,
+        )
+
+    def test_fit_partial_with_same_dataset(self, dataset: Dataset) -> None:
```
blondered, line 325: Please make this a test where factors are being compared. Same recommendations on small toy data  are not really a signal that `fit_partial` and `fit` are doing exactly the same logic. Please `fit` model for 10 epochs. Then `fit_partial` 10 times for one epoch on another copy of model. And then please compare that factors are the same

CHANGELOG.md
blondered: Thanks! Sorry for the delay. It's summer and it was a vacation time :)

chezou:
Thanks for dedicating your time and effort to my PR. Your support is invaluable and greatly appreciated.

I decided to close this PR because 
- There can still be substantial unwritten assumptions about user/item features that are unknown to me as an outsider. And it seems the gap is becoming a huge blocker to understand the intention of comments.
- Implicit ALS's user/item features handling and Dataset structure may need drastic changes to adopt warm feature handling.

My assumption for user/item features was:
- There should be user/item tables along with the interaction table.
- Those tables' columns are additive during the incremental training phase. If a column is removed, we can do full training.
- Looking at the [Lightfm's model resize workaround](https://github.com/lyst/lightfm/issues/347#issuecomment-707829342), it looks like features are additive, not dropped.

In our case, we mainly handle interactions only with Implicit. I'm glad to know some code in this PR works without features for ALS.

I know I could be better off joining your Telegram channel but installing another chat tool is not what I want.

PR #191, opened 2024-09-19
Feature/item net

In48semenov:
## Description

Added embeddings categorical item features to ItemNetBlock, Constructed Item Net for embeddings items ids and features. Also added method to retrieve categorical features from SparseFeatures.

## Type of change

- [ ] Optimization

## How Has This Been Tested?


In48semenov:
TODO:
- [ ] Tutorial notebook for SasRec model with category features
- [ ] DocString
- [ ] Tests

blondered (changes requested):

rectools/dataset/features.py
```diff
@@ -450,5 +450,21 @@ def __len__(self) -> int:
...
+        for idx, (name, value) in enumerate(self.names):
+            if value != DIRECT_FEATURE_VALUE:
+                feature_name: SparseFeatureName = (name, value)
+                cat_features_names_list.append(feature_name)
+                ids_cat_features_names.append(idx)
+
+        cat_features_names_tuple: tp.Tuple[SparseFeatureName, ...] = tuple(cat_features_names_list)
+        return SparseFeatures(
+            values=self.values[:, ids_cat_features_names],
+            names=cat_features_names_tuple,
```
blondered, line 466:
```suggestion
            names=self.names[cat_feature_ids],
```

rectools/dataset/features.py
```diff
@@ -450,5 +450,21 @@ def __len__(self) -> int:
         """Return number of objects."""
         return self.values.shape[0]
 
+    def get_cat_features(self) -> "SparseFeatures":
+        """Return `SparseFeatures` only with categorical features.""" ""
+        cat_features_names_list: tp.List[SparseFeatureName] = []
```
blondered, line 455: You really don't need it.
We can make this piece of code more clean. Always try to write as clean as possible and drop all unnecessary complications. This will make your code much easier to read and maintain.

rectools/dataset/features.py
```diff
@@ -450,5 +450,21 @@ def __len__(self) -> int:
...
+        ids_cat_features_names = []
```
blondered, line 456:
```suggestion
        cat_feature_ids = []
```

blondered (changes requested):

rectools/models/sasrec.py
```diff
@@ -579,11 +822,13 @@ def __init__(
         n_heads: int,
         dropout_rate: float,
         use_pos_emb: bool = True,
+        use_ids_emb: bool = True,
```
blondered, line 825: My thoughts on interface (open for discussion):
We can't add these flags because we want to be able to pass any number of any custom net blocks. So we can't make flags for all of the blocks.

rectools/models/sasrec.py
```diff
@@ -579,11 +822,13 @@ def __init__(
...
+        use_cat_item_features_emb: bool = False,
         loss: str = "softmax",
         verbose: int = 0,
         cpu_n_threads: int = 0,
         transformer_layers_type: tp.Type[TransformerLayersBase] = SasRecTransformerLayers,  # SASRec authors net
-        item_net_type: tp.Type[ItemNetBase] = IdEmbeddingsItemNet,  # item embeddings on ids
+        item_net_type: tp.Type[ItemNetBase] = ConstructedItemNetBlock,  # item embeddings on ids and cat features
```
blondered, line 831: My thoughts on interface (open for discussion):
How about we make instead one of these variants:
1. `item_net_constructor  = ItemNetConstructor(item_net_blocks=[IdEmbeddingsItemNet, CatFeaturesItemNet])`
2. `item_net_block_types = [IdEmbeddingsItemNet, CatFeaturesItemNet]`

`CatFeaturesItemNet` is what you've called `CatFeaturesEmebbedingsItemBlock`

`ItemNetConstructor` is what you've called `ConstructedItemNetBlock`. But the class should be universal and it doesn't have any `if` inside it's code. It just runs `from_dataset` methods of each block. And it summurizes all blocks `forward` results. Nothing else

blondered (changes requested):

rectools/models/sasrec.py
```diff
@@ -52,6 +60,119 @@ def forward(self, seqs: torch.Tensor, timeline_mask: torch.Tensor, attn_mask: to
         raise NotImplementedError()
 
 
+class CatFeatureEmebbedingsItem(nn.Module):
```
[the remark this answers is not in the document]
In48semenov: I think the `mapping feature names to their nets` functionality is handy at least for viewing the architecture of the whole model.
But I agree that `CatFeatureEmebbedingsItem` is an unnecessary net.
So I would suggest removing `CatFeatureEmebbedingsItem` and instead use the following code in `CatFeaturesEmebbedingsItemBlock` in `__init__` method:

`
self.category_embeddings = nn.ModuleDict(
    {
        feature_name: nn.Sequential(OrderedDict([
          ('value_emb', nn.Embedding(num_embeddings=feature_values.size, embedding_dim=n_factors)),
          ('drop_layer', nn.Dropout(dropout_rate)),
        ]))
        for feature_name, feature_values in common_name_features_ids_map.items()
    }
)
`

The forward method in `CatFeaturesEmebbedingsItemBlock` will have to be adjusted a bit. For a particular category feature, get all its embeddings, having previously stored how many unique values are in each category feature.

rectools/models/sasrec.py
In48semenov: If I understand you correctly, maybe make them optional? Since enabling and disabling the use of embeddings of item ids and category item features is a good feature in my opinion

rectools/models/sasrec.py
In48semenov: I did this now because I wanted to emphasis in the architecture of the net itself the names of the blocks that are used in the net. If it is not necessary, you can try one of the variants you suggested, but until it is hard to answer which one is better.

rectools/models/sasrec.py
blondered: We can enable and disable them just by passing `item_net_block_types = [IdEmbeddingsItemNet, CatFeaturesItemNet]` or `item_net_block_types = [IdEmbeddingsItemNet]`.
This will have the same functionality of switching different net blocks on and off but it would also allow any number of custom blocks to be passed.

In48semenov: Yes, I agree.

rectools/models/sasrec.py
```diff
@@ -52,6 +60,119 @@ def forward(self, seqs: torch.Tensor, timeline_mask: torch.Tensor, attn_mask: to
...
+    def catalogue_values(self) -> torch.Tensor:
+        """TODO"""
+        return torch.arange(0, self.n_values, device=self.device)
+
+    def get_all_embeddings(self) -> torch.Tensor:
+        """TODO"""
+        return self.forward(self.catalogue_values)
+
+    @classmethod
+    def from_array(cls, values: np.ndarray, n_factors: int, dropout_rate: float) -> tpe.Self:
```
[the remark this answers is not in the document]
blondered: `from_dataset` is a common method for all item net blocks. It is needed to treat every item net absolutely equal in the code. `IdEmbeddingsItemNet` will take only `dataset.item_id_map.size` but your `CategoryFeaturesItemNet` will take item category features. And other nets will take other dataset components. That was the main motivation for the method.
Here I don't really see motivation for `from_array`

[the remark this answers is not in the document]
In48semenov: I think, I will remove net for single category feature.

rectools/models/sasrec.py
```diff
@@ -462,10 +674,37 @@ def process_dataset_train(self, dataset: Dataset) -> Dataset:
...
+                values=item_features_values, names=item_features_names
+            )
+
         interactions = Interactions.from_raw(interactions, user_id_map, item_id_map)
-        dataset = Dataset(user_id_map, item_id_map, interactions)
+
+        dataset = Dataset(user_id_map, item_id_map, interactions, item_features=prepared_item_features)
 
         self.item_id_map = dataset.item_id_map
+        self.item_features = prepared_item_features
```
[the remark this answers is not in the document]
blondered: `transform_dataset_u2i` and `transform_dataset_i2i` are important because they are used afterwards in ModelBase code to create target users and items in internal ids. And user interactions are also passed to recommend dataloader.
But both methods use only id mapping, no actual data of previous interactions or features. Saving id map to the model during fitting is fine. Saving actual data is not good if it can be avoided.

rectools/dataset/features.py
```diff
@@ -450,5 +450,21 @@ def __len__(self) -> int:
...
+        for idx, (name, value) in enumerate(self.names):
+            if value != DIRECT_FEATURE_VALUE:
+                feature_name: SparseFeatureName = (name, value)
+                cat_features_names_list.append(feature_name)
+                ids_cat_features_names.append(idx)
+
+        cat_features_names_tuple: tp.Tuple[SparseFeatureName, ...] = tuple(cat_features_names_list)
```
feldlime, line 463: It's not the best practice to put var type into its name, you don't really need it here. Actually you don't really need this var at all

rectools/models/sasrec.py
```diff
@@ -462,10 +674,37 @@ def process_dataset_train(self, dataset: Dataset) -> Dataset:
...
+        # get item features
+        prepared_item_features = None
+        if dataset.item_features is not None:
+            item_features = dataset.item_features
+            if not isinstance(item_features, SparseFeatures):
+                raise ValueError("`item_features` in `dataset` must be `SparseFeatures` instance.")
+
+            item_features_names = item_features.names
+
+            internal_ids = dataset.item_id_map.convert_to_internal(interactions[Columns.Item].unique())
```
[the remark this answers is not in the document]
feldlime: Why not just `internal_ids = dataset.item_id_map.internal_ids[self.n_item_extra_tokens:]`?

Don't understand if you care about the order or not

rectools/models/sasrec.py
```diff
@@ -462,10 +674,37 @@ def process_dataset_train(self, dataset: Dataset) -> Dataset:
...
+            item_features_by_sorted_interactions = item_features.take(internal_ids)
+
+            dtype = item_features_by_sorted_interactions.values.dtype
+            num_features = item_features_by_sorted_interactions.values.shape[1]
+            pad_item_features = sp.csr_matrix((1, num_features), dtype=dtype)
```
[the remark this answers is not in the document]
feldlime: Let's please use `sparse` (not `sp`) since we do this everywhere in this library

rectools/models/sasrec.py
```diff
@@ -52,6 +60,119 @@ def forward(self, seqs: torch.Tensor, timeline_mask: torch.Tensor, attn_mask: to
...
+        feature_embeddings = self.drop_layer(feature_embeddings)
+        return feature_embeddings
+
+    @property
+    def device(self) -> torch.device:
+        """TODO"""
+        return self.value_emb.weight.device
+
+    @property
+    def catalogue_values(self) -> torch.Tensor:
```
blondered, line 85: This property will only be necessary for `ItemNetConstructor` class. Jther item net blocks do not need it at all.
Same for `get_all_embeddings`.

This comes because of our decision to accept `item_net_blocks: tp.List[tp.Type[ItemNetBlockBase]]` instead of `item_net_type: tp.Type[ItemNetBase]` in model interface.

blondered (changes requested):

rectools/models/sasrec.py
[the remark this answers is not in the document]
In48semenov: Not prepared dataset is used to get the item features, so it doesn't know anything about padding tokens.

Also, we need the order of item ids for correct compilation of item tokens.

rectools/dataset/features.py
In48semenov: I modified your suggestion. Since `self.names` is `Tuple` and `cat_feature_ids` is `List[int]`, but you can only get the required elements from Tuple object using slice and integer.

rectools/models/sasrec.py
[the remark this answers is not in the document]
blondered: I agree that `dataset.item_id_map.internal_ids[self.n_item_extra_tokens:]` will not work

rectools/models/sasrec.py
```diff
@@ -406,8 +545,32 @@ def process_dataset_train(self, dataset: Dataset) -> Dataset:
...
+                raise ValueError("`item_features` in `dataset` must be `SparseFeatures` instance.")
+
+            internal_ids = dataset.item_id_map.convert_to_internal(
+                item_id_map.get_external_sorted_by_internal()[self.n_item_extra_tokens :]
+            )
+            sorted_item_features = item_features.take(internal_ids)
+
+            dtype = sorted_item_features.values.dtype
+            n_features = sorted_item_features.values.shape[1]
+            pad_item_features = sparse.csr_matrix((self.n_item_extra_tokens, n_features), dtype=dtype)
```
blondered, line 563:
```suggestion
            extra_token_feature_values = sparse.csr_matrix((self.n_item_extra_tokens, n_features), dtype=dtype)
```
Naming must be logical. Always) It's not "pad" because there can be "mask" and any other extra tokens.
It's not "_features" because it is not SparseFeatures.

rectools/models/sasrec.py
```diff
@@ -406,8 +545,32 @@ def process_dataset_train(self, dataset: Dataset) -> Dataset:
...
+
+            extra_tokens_feature_values: sparse.scr_matrix = sparse.vstack(
```
blondered, line 565:
```suggestion
            full_feature_values: sparse.scr_matrix = sparse.vstack(
```
Naming must be logical. Always)

rectools/models/sasrec.py
```diff
@@ -406,8 +545,32 @@ def process_dataset_train(self, dataset: Dataset) -> Dataset:
...
+                [pad_item_features.toarray(), sorted_item_features.values], format="csr"
```
blondered, line 566: Do we need `toarray` here?

rectools/models/sasrec.py
```diff
@@ -406,8 +545,32 @@ def process_dataset_train(self, dataset: Dataset) -> Dataset:
         user_id_map = IdMap.from_values(interactions[Columns.User].values)
         item_id_map = IdMap.from_values(self.item_extra_tokens)
         item_id_map = item_id_map.add_ids(interactions[Columns.Item])
+
+        # get item features
+        item_features = None
+        if dataset.item_features is not None:
+            item_features = dataset.item_features
+            if not isinstance(item_features, SparseFeatures):
```
blondered, line 553: We don't need this check here because we can have item nets that support DenseFeatures and user might want to select them and provide DenseFeatures.

But right now we don''t have time to make the code universal for both Dense and Sparse features. So please write `#TODO: remove assumption on SparseFeatures and add Dense Features support`

rectools/models/sasrec.py
```diff
@@ -78,24 +153,85 @@ def __init__(self, n_factors: int, n_items: int, dropout_rate: float):
...
-        return torch.arange(0, self.n_items, device=self.item_emb.weight.device)
+        return torch.arange(0, self.n_items, device=self.device)
 
     def get_all_embeddings(self) -> torch.Tensor:
         """TODO"""
         return self.forward(self.catalogue)
 
     @classmethod
-    def from_dataset(cls, dataset: Dataset, n_factors: int, dropout_rate: float) -> tpe.Self:
+    def construct_nets_from_dataset(
```
blondered, line 219: I kind of like that you inherit `ItemNetConstructor` from `ItemNetBase`. Let's keep it up.
Let's rename `construct_nets_from_dataset` method to `from_dataset` and treat all of the following logic as a usual `ItemNetBase` `from_dataset` behaviour. It makes sense.

blondered (changes requested):

rectools/dataset/features.py
In48semenov: I agree. I removed that part of the code

blondered (approved):

PR #197, opened 2024-10-25
Bert4rec

spirinamayya:
Added bert4rec model

### DIFF
--- a/rectools/models/bert4rec.py
+++ b/rectools/models/bert4rec.py
@@ -57,7 +57,7 @@ class BERT4RecDataPreparator(SessionEncoderDataPreparatorBase):
                 if random_probs[j] < 0.8:
                     masked_session[j] = 1
                 elif random_probs[j] < 0.9:
-                    masked_session[j] = np.random.randint(low=2, high=self.item_id_map.size, size=1)[0]
+                    masked_session[j] = np.random.randint(low=self.n_item_extra_tokens, high=self.item_id_map.size)
             else:
                 target[j] = 0
         return masked_session, target
@@ -68,24 +68,24 @@ class BERT4RecDataPreparator(SessionEncoderDataPreparatorBase):
     ) -> Tuple[torch.LongTensor, torch.LongTensor, torch.FloatTensor]:
         """TODO"""
         batch_size = len(batch)
-        x = np.zeros((batch_size, self.session_max_len))
-        y = np.zeros((batch_size, self.session_max_len))
-        yw = np.zeros((batch_size, self.session_max_len))
+        x = np.zeros((batch_size, self.session_max_len + 1))
+        y = np.zeros((batch_size, self.session_max_len + 1))
+        yw = np.zeros((batch_size, self.session_max_len + 1))
         for i, (ses, ses_weights) in enumerate(batch):
             masked_session, target = self._mask_session(ses)
-            x[i, -len(ses) + 1 :] = masked_session[:-1]  # ses: [session_len] -> x[i]: [session_max_len]
-            y[i, -len(ses) + 1 :] = target[:-1]  # ses: [session_len] -> y[i]: [session_max_len]
-            yw[i, -len(ses) + 1 :] = ses_weights[:-1]  # ses_weights: [session_len] -> yw[i]: [session_max_len]
+            x[i, -len(ses) :] = masked_session  # ses: [session_len] -> x[i]: [session_max_len + 1]
+            y[i, -len(ses) :] = target  # ses: [session_len] -> y[i]: [session_max_len + 1]
+            yw[i, -len(ses) :] = ses_weights  # ses_weights: [session_len] -> yw[i]: [session_max_len + 1]
 
         return torch.LongTensor(x), torch.LongTensor(y), torch.FloatTensor(yw)
 
     def _collate_fn_recommend(self, batch: List[Tuple[List[int], List[float]]]) -> torch.LongTensor:
         """Right truncation, left padding to session_max_len"""
-        x = np.zeros((len(batch), self.session_max_len))
+        x = np.zeros((len(batch), self.session_max_len + 1))
         for i, (ses, _) in enumerate(batch):
             session = ses.copy()
-            session = session[1:] + [1]
-            x[i, -len(ses) :] = session[-self.session_max_len :]
+            session = session + [1]
+            x[i, -len(ses) - 1 :] = session[-self.session_max_len - 1 :]
         return torch.LongTensor(x)
 
 
@@ -132,7 +132,8 @@ class BERT4RecTransformerLayers(TransformerLayersBase):
             ff_output = self.feed_forward[i](ff_input)
             seqs = seqs + self.dropout2[i](ff_output)
             seqs = self.dropout3[i](seqs)
-
+        # TODO: test with torch.nn.Linear and cross-entropy loss as in
+        # https://github.com/jaywonchung/BERT4Rec-VAE-Pytorch/blob/f66f2534ebfd937778c7174b5f9f216efdebe5de/models/bert.py#L11C1-L11C67
         return seqs
 
 
@@ -146,7 +147,7 @@ class BERT4RecModel(TransformerModelBase):
         n_factors: int = 128,
         use_pos_emb: bool = True,
         use_causal_attn: bool = False,
-        use_mlm_attn: bool = True,
+        use_key_padding_mask: bool = True,
         dropout_rate: float = 0.2,
         epochs: int = 3,
         verbose: int = 0,
@@ -168,12 +169,13 @@ class BERT4RecModel(TransformerModelBase):
     ):
         super().__init__(
             transformer_layers_type=transformer_layers_type,
+            data_preparator_type=data_preparator_type,
             n_blocks=n_blocks,
             n_heads=n_heads,
             n_factors=n_factors,
             use_pos_emb=use_pos_emb,
             use_causal_attn=use_causal_attn,
-            use_mlm_attn=use_mlm_attn,
+            use_key_padding_mask=use_key_padding_mask,
             dropout_rate=dropout_rate,
             epochs=epochs,
             verbose=verbose,
@@ -181,7 +183,7 @@ class BERT4RecModel(TransformerModelBase):
             cpu_n_threads=cpu_n_threads,
             loss=loss,
             lr=lr,
-            session_max_len=session_max_len,
+            session_max_len=session_max_len + 1,
             trainer=trainer,
             item_net_block_types=item_net_block_types,
             pos_encoding_type=pos_encoding_type,
--- a/rectools/models/bert4rec.py
+++ b/rectools/models/bert4rec.py
@@ -1,35 +1,23 @@
 import typing as tp
-import warnings
-from copy import deepcopy
 from typing import List, Tuple
 
 import numpy as np
 import torch
-import tqdm
 from pytorch_lightning import Trainer
-from scipy import sparse
 from torch import nn
-from torch.utils.data import DataLoader
 
-from rectools import Columns, ExternalIds
-from rectools.dataset import Dataset, Interactions
-from rectools.dataset.features import SparseFeatures
-from rectools.dataset.identifiers import IdMap
-from rectools.models.base import ErrorBehaviour, InternalRecoTriplet, ModelBase
-from rectools.models.rank import Distance, ImplicitRanker
 from rectools.models.sasrec import (
     CatFeaturesItemNet,
     IdEmbeddingsItemNet,
     ItemNetBase,
-    ItemNetConstructor,
     LearnableInversePositionalEncoding,
     PositionalEncodingBase,
-    SequenceDataset,
     SessionEncoderDataPreparatorBase,
+    SessionEncoderLightningModule,
     SessionEncoderLightningModuleBase,
     TransformerLayersBase,
+    TransformerModelBase,
 )
-from rectools.types import InternalIdsArray
 
 PADDING_VALUE = "PAD"
 MASKING_VALUE = "MASK"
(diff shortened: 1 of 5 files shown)


rectools/models/bert4rec.py
```diff
@@ -0,0 +1,578 @@
...
+        verbose: int = 0,
+        deterministic: bool = False,
+        cpu_n_threads: int = 0,
+        session_max_len: int = 32,
+        batch_size: int = 128,
+        loss: str = "softmax",
+        lr: float = 0.01,
+        dataloader_num_workers: int = 0,
+        train_min_user_interaction: int = 2,
+        mask_prob: float = 0.15,
```
blondered, line 424: As far as I can see the only difference in this model from SASRec is receiving mask_prob and passing it to data_preparator_type.
What you need to do:
1. Create `TransformerModelBase` class with all of the common methods already implemented. Look here https://github.com/MobileTeleSystems/RecTools/blob/main/rectools/models/base.py for example. Carefully select which of the arguments must be passed to __init__. `mask_prob` shouldn't be there at all. `transformer_layers_type` and `data_preparator_type` should not have default values. `self.data_preparator` should not be initialized but should be declared like `self.data_preparator: SessionEncoderDataPreparatorBase`. 
2. In SASRecModel __init__ create default values for `transformer_layers_type` and `data_preparator_type`. Initialize self.data_preparator
3. In BERT4RecModel __init__ create default values for `transformer_layers_type` and `data_preparator_type`. Initialize self.data_preparator

(This is a draft for now. I will update it soon. Don't implement for now)

blondered (changes requested):

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,578 @@
...
+            sequence_dataset,
+            batch_size=self.batch_size,
+            collate_fn=self._collate_fn_recommend,
+            num_workers=self.dataloader_num_workers,
+            shuffle=False,
+        )
+        return recommend_dataloader
+
+
+class PointWiseFeedForward(nn.Module):
```
blondered, line 221: Now about the feedforward differences.
There are a few:
1. The order of dropout and activation. For RELU it doesn't have any difference in the result. So let's change the order for SASRec variant and it will be the same as BERT4Rec.
2. Activation function. Let's just add and argument to PointWiseFeedForward __init__ which activation to use
3. Placing second dropout inside PointWiseFeedForward.forward or moving it to TransformerLayers.forward. We can move them to TransformerLayers in SASRecTransformerLayers so that it would be more like the classic picture of transformers architecture.

As far as I can see after all of this we will have one PointWiseFeedForward class for both models. And it's good.
Please make this change as a separate commit and check SASRec metrics.

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,578 @@
...
+        )
+        self.dropout2 = nn.ModuleList([nn.Dropout(dropout_rate) for _ in range(n_blocks)])
+        # self.dropout3 = nn.ModuleList([nn.Dropout(dropout_rate) for _ in range(n_blocks)])
+
+    def forward(self, seqs: torch.Tensor, timeline_mask: torch.Tensor, attn_mask: torch.Tensor) -> torch.Tensor:
+        """TODO"""
+        for i in range(self.n_blocks):
+            mha_input = self.layer_norm1[i](seqs)
+            # mha_output, _ =
+            # self.multi_head_attn[i](mha_input, mha_input, mha_input, attn_mask=attn_mask, need_weights=False)
```
blondered, line 268: This is not handy at all :)
Here just pass the one you received in `forward` always.
And create a flag in the model if you want to create this mask or not. If not - pass it with Null value

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,578 @@
...
+            mha_output, _ = self.multi_head_attn[i](mha_input, mha_input, mha_input, need_weights=False)
+            seqs = seqs + self.dropout1[i](mha_output)
+            ff_input = self.layer_norm2[i](seqs)
+            ff_output = self.feed_forward[i](ff_input)
+            seqs = seqs + self.dropout2[i](ff_output)
+            seqs = seqs * timeline_mask
+            # seqs = self.dropout3[i](seqs)
```
blondered, line 275: We don't need it, do we? Is there a reference for this?

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,578 @@
...
+        attn_mask = None
+        if self.use_causal_attn:
+            attn_mask = ~torch.tril(
+                torch.ones((session_max_len, session_max_len), dtype=torch.bool, device=sessions.device)
+            )
+        timeline_mask = sessions != 0
+        attn_mask = ~timeline_mask.unsqueeze(1).repeat(self.n_heads, timeline_mask.squeeze(-1).shape[1], 1)
+        timeline_mask = timeline_mask.unsqueeze(-1)
+        seqs = item_embs[sessions]  # [batch_size, session_max_len, n_factors]
+        seqs = self.pos_encoding(seqs, timeline_mask)
```
blondered, line 342: Decided to drop passing timeline_mask to pos_encoding

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,578 @@
...
+
+        """
+        session_max_len = sessions.shape[1]
+        attn_mask = None
+        if self.use_causal_attn:
+            attn_mask = ~torch.tril(
+                torch.ones((session_max_len, session_max_len), dtype=torch.bool, device=sessions.device)
+            )
+        timeline_mask = sessions != 0
+        attn_mask = ~timeline_mask.unsqueeze(1).repeat(self.n_heads, timeline_mask.squeeze(-1).shape[1], 1)
```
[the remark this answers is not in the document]
blondered: Decided to pass key_padding_mask instead of combining both masks in attn_mask

rectools/models/bert4rec.py
blondered: Also here for BERT we need to pass `~timeline_mask` as `key_padding_mask`

blondered (changes requested):

blondered:
To make common positional encoding let's drop multiplying on timeline_mask from it. For sasrec let's move it to SASRectransformerLayers, just before going to the layers

blondered:
Let's add "TODO: init padding embedding with zeros" to weights initialization code

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,578 @@
...
+        self.session_max_len = session_max_len
+        self.batch_size = batch_size
+        self.dataloader_num_workers = dataloader_num_workers
+        self.train_min_user_interactions = train_min_user_interactions
+        self.item_extra_tokens = item_extra_tokens
+        self.mask_prob = mask_prob
+        self.shuffle_train = shuffle_train
+        # TODO: add SequenceDatasetType for fit and recommend
+
+    def process_dataset_train(self, dataset: Dataset) -> Dataset:
```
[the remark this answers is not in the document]
spirinamayya: fixed

rectools/models/bert4rec.py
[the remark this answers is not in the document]
spirinamayya: fixed

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,578 @@
...
+        logits = session_embs @ item_embs.T  # [batch_size, session_max_len, n_items + 2]
+        return logits
+
+
+class SessionEncoderLightningModule(SessionEncoderLightningModuleBase):
+    """TODO"""
+
+    def on_train_start(self) -> None:
+        """TODO"""
+        self._truncated_normal_init()
```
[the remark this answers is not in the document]
spirinamayya: fixed

rectools/models/bert4rec.py
spirinamayya: fixed

rectools/models/bert4rec.py
spirinamayya: fixed

rectools/models/bert4rec.py
spirinamayya: fixed

rectools/models/bert4rec.py
spirinamayya: fixed

rectools/models/sasrec.py
```diff
@@ -1260,8 +1123,8 @@ def _recommend_i2i(
 
         ranker = ImplicitRanker(
             self.i2i_dist,
-            item_embs,  # [n_items + 1, n_factors]
-            item_embs,  # [n_items + 1, n_factors]
+            item_embs,  # [n_items + 2, n_factors]
```
blondered, line 1126:
```suggestion
            item_embs,  # [n_items + n_special_tokens, n_factors]
```
here and everywhere

blondered (changes requested):

rectools/models/sasrec.py
```diff
@@ -1066,67 +977,20 @@ def _xavier_normal_init(self) -> None:
...
-        Type of positional encoding.
-    transformer_layers_type: Type(TransformerLayersBase), default `SasRecTransformerLayers`
-        Type of transformer layers architecture.
-    data_preparator_type: Type(SessionEncoderDataPreparatorBase), default `SasRecDataPreparator`
-        Type of data preparator used for dataset processing and dataloader creation.
-    lightning_module_type: Type(SessionEncoderLightningModuleBase), default `SessionEncoderLightningModule`
-        Type of lightning module defining training procedure.
-    """
+class TransformerModelBase(ModelBase):
+    """TODO"""
```
blondered, line 981: """Base model for all recommender algorithms that work on transformer architecture (e.g. SASRec, Bert4Rec). To create a custom transformer model it is necessary to inherit from this class and write `self.data_preparator` initialization logic.
"""

blondered (changes requested):

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,197 @@
...
+        masked_session = ses.copy()
+        target = ses.copy()
+        random_probs = np.random.rand(len(ses))
+        for j in range(len(ses)):
+            if random_probs[j] < self.mask_prob:
+                random_probs[j] /= self.mask_prob
+                if random_probs[j] < 0.8:
+                    masked_session[j] = 1
+                elif random_probs[j] < 0.9:
+                    masked_session[j] = np.random.randint(low=2, high=self.item_id_map.size, size=1)[0]
```
blondered, line 60: low=self.n_item_extra_tokens

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def _mask_session(self, ses: List[int]) -> Tuple[List[int], List[int]]:
+        masked_session = ses.copy()
+        target = ses.copy()
+        random_probs = np.random.rand(len(ses))
+        for j in range(len(ses)):
+            if random_probs[j] < self.mask_prob:
+                random_probs[j] /= self.mask_prob
+                if random_probs[j] < 0.8:
+                    masked_session[j] = 1
```
blondered, line 58: Let's make it more clear and safe.
In the beginning:
mask_id = self.item_id_map.convert_to_internal([MASKING_VALUE])[0]
Here:
masked_session[j] = mask_id

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,197 @@
...
+        batch: List[Tuple[List[int], List[float]]],
+    ) -> Tuple[torch.LongTensor, torch.LongTensor, torch.FloatTensor]:
+        """TODO"""
+        batch_size = len(batch)
+        x = np.zeros((batch_size, self.session_max_len))
+        y = np.zeros((batch_size, self.session_max_len))
+        yw = np.zeros((batch_size, self.session_max_len))
+        for i, (ses, ses_weights) in enumerate(batch):
+            masked_session, target = self._mask_session(ses)
+            x[i, -len(ses) + 1 :] = masked_session[:-1]  # ses: [session_len] -> x[i]: [session_max_len]
```
blondered, line 76: Why are we not taking the last token in the seq to training?

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,197 @@
...
+            yw[i, -len(ses) + 1 :] = ses_weights[:-1]  # ses_weights: [session_len] -> yw[i]: [session_max_len]
+
+        return torch.LongTensor(x), torch.LongTensor(y), torch.FloatTensor(yw)
+
+    def _collate_fn_recommend(self, batch: List[Tuple[List[int], List[float]]]) -> torch.LongTensor:
+        """Right truncation, left padding to session_max_len"""
+        x = np.zeros((len(batch), self.session_max_len))
+        for i, (ses, _) in enumerate(batch):
+            session = ses.copy()
+            session = session[1:] + [1]
```
blondered, line 87: What is sessions had only two items? Why are we dropping the first one on inference?

rectools/models/bert4rec.py
```diff
@@ -0,0 +1,197 @@
...
+            )
+            seqs = seqs + self.dropout1[i](mha_output)
+            ff_input = self.layer_norm2[i](seqs)
+            ff_output = self.feed_forward[i](ff_input)
+            seqs = seqs + self.dropout2[i](ff_output)
+            seqs = self.dropout3[i](seqs)
+
+        return seqs
+
+
```
blondered, line 138: Let's add:
`# TODO: test with torch.nn.Linear and cross-entropy loss as in https://github.com/jaywonchung/BERT4Rec-VAE-Pytorch/blob/f66f2534ebfd937778c7174b5f9f216efdebe5de/models/bert.py#L11C1-L11C67`

spirinamayya: fixed

rectools/models/sasrec.py
```diff
@@ -1066,67 +977,20 @@ def _xavier_normal_init(self) -> None:
...
 
     def __init__(  # pylint: disable=too-many-arguments
         self,
+        transformer_layers_type: tp.Type[TransformerLayersBase],
```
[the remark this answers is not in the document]
spirinamayya: fixed

rectools/models/sasrec.py
```diff
@@ -518,6 +455,7 @@ def __init__(
         dropout_rate: float,
         use_pos_emb: bool = True,
         use_causal_attn: bool = True,
+        use_mlm_attn: bool = False,
```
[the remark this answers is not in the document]
spirinamayya: fixed

rectools/models/sasrec.py
spirinamayya: fixed

rectools/models/sasrec.py
spirinamayya: fixed

rectools/models/bert4rec.py
spirinamayya: fixed

rectools/models/bert4rec.py
spirinamayya: fixed

rectools/models/bert4rec.py
spirinamayya: fixed

rectools/models/bert4rec.py
spirinamayya: Training is significantly slower

rectools/models/sasrec.py
```diff
@@ -572,15 +512,19 @@ def encode_sessions(self, sessions: torch.Tensor, item_embs: torch.Tensor) -> to
...
             )
+        if self.use_key_padding_mask:
+            key_padding_mask = sessions == 0
         timeline_mask = (sessions != 0).unsqueeze(-1)  # [batch_size, session_max_len, 1]
         seqs = item_embs[sessions]  # [batch_size, session_max_len, n_factors]
-        seqs = self.pos_encoding(seqs, timeline_mask)
+        seqs = self.pos_encoding(seqs)
         seqs = self.emb_dropout(seqs)
-        seqs = self.transformer_layers(seqs, timeline_mask, attn_mask)
+        seqs = self.transformer_layers(seqs, timeline_mask, attn_mask, key_padding_mask)
```
blondered, line 527: Let's add # TODO: stop passing  timeline_mask together with key_padding_mask because they have same information

rectools/models/bert4rec.py
spirinamayya: fixed

rectools/models/sasrec.py
spirinamayya: fixed

Issue #199, opened 2024-10-30
recommend() method for ImplicitALSWrapperModel


Waujito:
### Your Question

Hello!
I see that here is no usage of ImplicitALSWrapperModel's recommend() method. I see that rectools' recommend() fallbacks to something underlying called ImplicitRanker with custom implementation by RecTools. I wonder why is this behaviour used instead of the native implicit's recommend(). Also this implementation works a way longer than implicit's.

blondered:
Hi!
ImplicitRanker uses exactly the same method for scoring items as in implicit library, we are calling it directly: https://github.com/MobileTeleSystems/RecTools/blob/0be5e1592e06022109faf1d446c7aee82fbd9dfb/rectools/models/rank.py#L194

In implicit framework `recommend` just returns the results directly without any processing. Items in recommendations can still include already-seen items and repeated items. They will get neginf scores, but they are still present.

In our wrapper we process those results, apply filtering based on scores here: https://github.com/MobileTeleSystems/RecTools/blob/0be5e1592e06022109faf1d446c7aee82fbd9dfb/rectools/models/rank.py#L106
And after that we convert results to pandas.

So this is why our `recommend` is slower: we do some post-processing.

Waujito:
Yes, thank you. I see it works the same way and the speed should be the same. My speed problem was because the RecTools' ImplicitRanker implementation does not use the GPU. I was able to run predictions on the GPU with the patch:
```diff
--- a/rank.py 2024-10-30 16:03:51.376808529 +0300
+++ b/rank.py     2024-10-30 16:07:06.556808606 +0300
@@ -18,6 +18,7 @@
 from enum import Enum
 
 import implicit.cpu
+import implicit.gpu
 import numpy as np
 from implicit.cpu.matrix_factorization_base import _filter_items_from_sparse_matrix as filter_items_from_sparse_matrix
 from scipy import sparse
@@ -191,14 +192,22 @@
 
         real_k = min(k, object_factors.shape[0])
 
-        ids, scores = implicit.cpu.topk.topk(  # pylint: disable=c-extension-no-member
-            items=object_factors,
-            query=subject_factors,
-            k=real_k,
+        object_factors = implicit.gpu.Matrix(object_factors)
+        subject_factors = implicit.gpu.Matrix(subject_factors)
+        if object_norms is not None:
+            object_norms = implicit.gpu.Matrix(object_norms)
+        if filter_query_items is not None:
[...]

PR #201, opened 2024-11-01
GPU support for ImplicitRanker

### LINKED ISSUE #200: gpu support for `ImplicitRanker`
### Feature Description

`use_gpu` argument to `ImplicitRanker.rank` method that calls `implicit.gpu.KnnQuery().topk` instead of `implicit.cpu.topk.topk` when set to True.

### Why this feature?

Allows for gpu inference for a range of models

### Additional context

Described at https://github.com/MobileTeleSystems/RecTools/issues/199

Waujito:
## Type of change

## Description

Implement use_gpu argument to ImplicitRanker.rank method. Adapt tests to it.

This PR closes #200  and #199 .

The GPU support works well but it does not pass all the test normally. The reason is behind implicit's implementation of topk on GPU and is not a bug: if predicted scores are equal, the order of elements is undefined. On CPU it is sorted in an ascending order but on GPU the order seems to be reverse. This *limitation* fails a lot of tests. Im not sure how to handle this because here are too much test cases to be rewritten without this thing.

Also I should mention changes in _get_neginf_score. On GPU the filtration score is slightly different than on CPU:  https://github.com/benfred/implicit/blob/main/implicit/gpu/knn.cu#L36 
I changed the implementation and its tests.

So, questions like interface for user, GPU by default, tests for ranker on GPU keep open and feedback from mainterers will be fine.

PS.
For testers: make sure implicit has GPU support on specific environment(`python -c "from implicit.gpu import HAS_CUDA; print(HAS_CUDA)"`). It is really unstable thing and I was able to run it normally only in the custom environment. On one created with `make install` it was falling back to disabled GPU.

### DIFF
--- a/rectools/models/rank.py
+++ b/rectools/models/rank.py
@@ -15,11 +15,14 @@
 """Implicit ranker model."""
 
 import typing as tp
+import warnings
 from enum import Enum
 
 import implicit.cpu
+import implicit.gpu
 import numpy as np
 from implicit.cpu.matrix_factorization_base import _filter_items_from_sparse_matrix as filter_items_from_sparse_matrix
+from implicit.gpu import HAS_CUDA
 from scipy import sparse
 
 from rectools import InternalIds
@@ -74,8 +77,14 @@ class ImplicitRanker:
             self.subjects_dots = self._calc_dots(self.subjects_factors)
 
     def _get_neginf_score(self) -> float:
-        # Adding 1 to avoid float calculation errors (we're comparing `scores <= neginf_score`)
-        return float(-np.finfo(np.float32).max + 1)
+        # neginf_score computed according to implicit gpu FLT_FILTER_DISTANCE
+        # https://github.com/benfred/implicit/blob/main/implicit/gpu/knn.cu#L36
+        # we're comparing `scores <= neginf_score`
+        return float(
+            np.asarray(
+                np.asarray(-np.finfo(np.float32).max, dtype=np.float32).view(np.uint32) - 1, dtype=np.uint32
+            ).view(np.float32)
+        )
 
     @staticmethod
     def _calc_dots(factors: np.ndarray) -> np.ndarray:
@@ -132,13 +141,50 @@ class ImplicitRanker:
 
         return all_target_ids, np.concatenate(all_reco_ids), np.concatenate(all_scores)
 
-    def rank(
+    def _rank_on_gpu(
+        self,
+        object_factors: np.ndarray,
+        subject_factors: tp.Union[np.ndarray, sparse.csr_matrix],
+        k: int,
+        object_norms: tp.Optional[np.ndarray],
+        filter_query_items: tp.Optional[tp.Union[sparse.csr_matrix, sparse.csr_array]],
+    ) -> tp.Tuple[np.ndarray, np.ndarray]:  # pragma: no cover
+        object_factors = implicit.gpu.Matrix(object_factors.astype(np.float32))
+
+        if isinstance(subject_factors, sparse.spmatrix):
+            warnings.warn("Sparse subject factors converted to Dense matrix")
+            subject_factors = subject_factors.todense()
+
+        subject_factors = implicit.gpu.Matrix(subject_factors.astype(np.float32))
+
+        if object_norms is not None:
+            if len(np.shape(object_norms)) == 1:
+                object_norms = np.expand_dims(object_norms, axis=0)
+            object_norms = implicit.gpu.Matrix(object_norms)
+
+        if filter_query_items is not None:
+            filter_query_items = implicit.gpu.COOMatrix(filter_query_items.tocoo())
+
+        ids, scores = implicit.gpu.KnnQuery().topk(  # pylint: disable=c-extension-no-member
+            items=object_factors,
+            m=subject_factors,
+            k=k,
+            item_norms=object_norms,
+            query_filter=filter_query_items,
+            item_filter=None,
+        )
+
+        scores = scores.astype(np.float64)
+        return ids, scores
+
+    def rank(  # pylint: disable=too-many-branches
         self,
         subject_ids: InternalIds,
         k: int,
         filter_pairs_csr: tp.Optional[sparse.csr_matrix] = None,
         sorted_object_whitelist: tp.Optional[InternalIdsArray] = None,
         num_threads: int = 0,
+        use_gpu: bool = False,
     ) -> tp.Tuple[InternalIds, InternalIds, Scores]:
         """Rank objects to proceed inference using implicit library topk cpu method.
 
@@ -156,7 +202,9 @@ class ImplicitRanker:
             If given, only these items will be used for recommendations.
             Otherwise all items from dataset will be used.
         num_threads : int, default 0
-            Will be used as `num_threads` parameter for `implicit.cpu.topk.topk`.
+            Will be used as `num_threads` parameter for `implicit.cpu.topk.topk`. Omitted if use_gpu is True
+        use_gpu : bool, default False
+            If True `implicit.gpu.KnnQuery().topk` will be used instead of classic cpu version.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


Waujito:
Force-pushed with linter fix.

blondered (commented):
Thank you for the PR! I'm suggesting some changes to the test inputs so that we don't have to handle different results from cpu and gpu ranking.

rectools/models/rank.py
```diff
@@ -191,15 +203,50 @@ def rank(
...
-            items=object_factors,
-            query=subject_factors,
-            k=real_k,
-            item_norms=object_norms,  # query norms for COSINE distance are applied afterwards
-            filter_query_items=filter_query_items,  # queries x objects csr matrix for getting neginf scores
-            filter_items=None,  # rectools doesn't support blacklist for now
-            num_threads=num_threads,
-        )
+        if use_gpu and not HAS_CUDA:
+            warnings.warn("Forced rank() on CPU")
```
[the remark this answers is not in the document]
Waujito: If I understand you right, I should keep these lines and remove HAS_CUDA guards in tests?
```py
@pytest.mark.parametrize("use_gpu", (False, True) if HAS_CUDA else (False,))
```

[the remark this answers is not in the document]
blondered: Yes, let's just do `@pytest.mark.parametrize("use_gpu", (False, True))`. Because you wrote a fallback behaviour and now we expect same results from cpu and gpu on all tests.

tests/models/test_rank.py
```diff
@@ -108,14 +110,18 @@ def test_rank(
...
             subject_factors = sparse.csr_matrix(subject_factors)
 
         ranker = ImplicitRanker(distance, subject_factors, object_factors)
-        _, actual_recs, actual_scores = ranker.rank(subject_ids=[0, 1], k=3)
-        np.testing.assert_equal(actual_recs, expected_recs)
+        _, actual_recs, actual_scores = ranker.rank(subject_ids=[0, 1], k=3, use_gpu=use_gpu)
         np.testing.assert_almost_equal(actual_scores, expected_scores)
+        if not use_gpu:
+            np.testing.assert_equal(actual_recs, expected_recs)
+        else:
```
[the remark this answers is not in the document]
blondered: This is strange, I don't see a pattern for these exact models.
I will look into this. Please give more details: what are you changing in tests and what do you get (just one example is enough).
We can't run gpu tests on github but I will run them locally.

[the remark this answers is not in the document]
Waujito:
```diff
diff --git a/rectools/models/rank.py b/rectools/models/rank.py
index bc43f12..6ea682f 100644
--- a/rectools/models/rank.py
+++ b/rectools/models/rank.py
@@ -148,7 +148,7 @@ class ImplicitRanker:
         filter_pairs_csr: tp.Optional[sparse.csr_matrix] = None,
         sorted_object_whitelist: tp.Optional[InternalIdsArray] = None,
         num_threads: int = 0,
-        use_gpu: bool = False,
+        use_gpu: bool = True,
     ) -> tp.Tuple[InternalIds, InternalIds, Scores]:
         """Rank objects to proceed inference using implicit library topk cpu method.
```

leads to 
```
_ TestVectorModel.test_without_biases[u2i-Distance.DOT-expected_reco0-expected_scores0] _

self = <tests.models.test_vector.TestVectorModel object at 0x7e9ad2873bc0>
distance = <Distance.DOT: 1>
expected_reco = [[0, 1, 2], [2, 1, 0]]
expected_scores = [[25, 0, -1], [0, 0, 0]], method = 'u2i'
...
>       assert list(reco) == sum(expected_reco, [])
E       assert [0, 1, 2, 0, 1, 2] == [0, 1, 2, 2, 1, 0]
E
E         At index 3 diff: 0 != 2
E         Use -v to get more diff

tests/models/test_vector.py:95: AssertionError
```

```
FAILED tests/models/test_vector.py::TestVectorModel::test_without_biases[u2i-Distance.DOT-expected_reco0-expected_scores0] - assert [0, 1, 2, 0, 1, 2] == [0, 1, 2, 2, 1, 0]
FAILED tests/models/test_vector.py::TestVectorModel::test_without_biases[u2i-Distance.COSINE-expected_reco1-expected_scores1] - assert [0, 1, 2, 0, 1, 2] == [0, 1, 2, 2, 1, 0]
FAILED tests/models/test_vector.py::TestVectorModel::test_without_biases[i2i-Distance.DOT-expected_reco0-expected_scores0] - assert [0, 1, 2, 0, 1, 2] == [0, 1, 2, 2, 1, 0]
FAILED tests/models/test_vector.py::TestVectorModel::test_without_biases[i2i-Distance.COSINE-expected_reco1-expected_scores1] - assert [0, 1, 2, 0, 1, 2] == [0, 1, 2, 2, 1, 0]
```

That's not a problem now, but may lead to problems with tests when use_gpu will be used as default. So, tests for these models should be rewritten as well to escape undefined behaviour.

rectools/models/rank.py
```diff
@@ -156,7 +166,9 @@ def rank(
             If given, only these items will be used for recommendations.
             Otherwise all items from dataset will be used.
         num_threads : int, default 0
-            Will be used as `num_threads` parameter for `implicit.cpu.topk.topk`.
+            Will be used as `num_threads` parameter for `implicit.cpu.topk.topk`. Omitted if use_gpu is True
+        use_gpu : bool, default False
```
blondered, line 170: Please update changelog: "Added `use_gpu` argument to `ImplicitRanker.rank` method"

Waujito: Done

rectools/models/rank.py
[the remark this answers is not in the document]
Waujito: Done

tests/models/test_rank.py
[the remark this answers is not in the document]
blondered: Thank you! We will process this in the next PR tests when we enable `use_gpu` for models.

rectools/models/rank.py
```diff
@@ -191,15 +203,47 @@ def rank(
...
-            item_norms=object_norms,  # query norms for COSINE distance are applied afterwards
-            filter_query_items=filter_query_items,  # queries x objects csr matrix for getting neginf scores
-            filter_items=None,  # rectools doesn't support blacklist for now
-            num_threads=num_threads,
-        )
+        if use_gpu and not HAS_CUDA:
+            warnings.warn("Forced rank() on CPU")
+            use_gpu = False
+
+        if use_gpu:
```
blondered, line 210:
```suggestion
        if use_gpu:  # pragma: no cover
```
To stop coverage from failing since we don't have gpu for tests on github.

Also please move all gpu logic to a separate private method. It makes sense and linters will stop complaining.

We merge as soon as all tests pass.

Waujito: I hope now it will run without any problem. Also I noticed that I forgot to add tests for sparse subject factors, fixed.

blondered: Please add `# pragma: no cover` comment to ignore gpu logic for coverage test.

rectools/models/rank.py
```diff
@@ -132,13 +141,50 @@ def _process_implicit_scores(
...
+            m=subject_factors,
+            k=k,
+            item_norms=object_norms,
+            query_filter=filter_query_items,
+            item_filter=None,
+        )
+
+        scores = scores.astype(np.float64)
+        return ids, scores
+
```
blondered, line 179: Please add `# pylint: disable=too-many-branches` comment next to `def rank(`

We do have a bit complicated logic in this method but we will refactor it later. So let's just ignore for now.

rectools/models/rank.py
Waujito: Done

rectools/models/rank.py
Waujito: Done

blondered (approved):
Thank you for contributing! We are preparing for the upcoming release now.
You can enable `use_gpu` for iALS in the next PR if you like. The logic for this model is very simple: use_gpu should be set to True if wrapped model is an instance of GPUAlternatingLeastSquares.

For other models we need a bit of thinking. I'm not sure if we are going to use gpu by default. The alternative is to make it an option for user to decide. Anyway, it's gonna take some time to decide and implement (and fix tests).

Waujito:
Nice! Thank you for RecTools!

PR #202, opened 2024-11-02
ItemNet tests and docs

In48semenov:
## Type of change

- [ ] Optimization

## How Has This Been Tested?

### DIFF
--- a/rectools/models/sasrec.py
+++ b/rectools/models/sasrec.py
@@ -89,7 +89,7 @@ class CatFeaturesItemNet(ItemNetBase):
 
     def forward(self, items: torch.Tensor) -> torch.Tensor:
         """TODO"""
-        # TODO: Should we use torch.nn.EmbeddingBag.html?
+        # TODO: Should we use torch.nn.EmbeddingBag?
         feature_dense = self.get_dense_item_features(items)
 
         feature_embs = self.category_embeddings(self.feature_catalogue)
--- a/rectools/models/sasrec.py
+++ b/rectools/models/sasrec.py
@@ -46,7 +46,7 @@ class ItemNetBase(nn.Module):
 
     @property
     def device(self) -> torch.device:
-        """TODO"""
+        """Return device the item net is located."""
         raise NotImplementedError()
 
 
@@ -68,8 +68,16 @@ class PositionalEncodingBase(torch.nn.Module):
 
 class CatFeaturesItemNet(ItemNetBase):
     """
-    Base class for all category item features embeddings. To use more complicated logic then just id embeddings inherit
-    from this class and pass your custom ItemNet to your model params.
+    Network for item embeddings based only on categorical item features.
+
+    Parameters
+    ----------
+    item_features: SparseFeatures
+        Storage for sparse features.
+    n_factors: int
+        Latent embedding size of item embeddings.
+    dropout_rate: float
+        Probability of a hidden unit to be zeroed.
     """
 
     def __init__(
@@ -88,7 +96,19 @@ class CatFeaturesItemNet(ItemNetBase):
         self.drop_layer = nn.Dropout(dropout_rate)
 
     def forward(self, items: torch.Tensor) -> torch.Tensor:
-        """TODO"""
+        """
+        Forward pass to get item embeddings from categorical item features.
+
+        Parameters
+        ----------
+        items: torch.Tensor
+            Internal item ids.
+
+        Returns
+        -------
+        torch.Tensor
+            Item embeddings.
+        """
         # TODO: Should we use torch.nn.EmbeddingBag?
         feature_dense = self.get_dense_item_features(items)
 
@@ -100,23 +120,46 @@ class CatFeaturesItemNet(ItemNetBase):
 
     @property
     def device(self) -> torch.device:
-        """TODO"""
+        """Return device the item net is located."""
         return self.category_embeddings.weight.device
 
     @property
     def feature_catalogue(self) -> torch.Tensor:
-        """TODO"""
+        """Return tensor with elements in range [0, n_cat_features)."""
         return torch.arange(0, self.n_cat_features, device=self.device)
 
     def get_dense_item_features(self, items: torch.Tensor) -> torch.Tensor:
-        """TODO"""
+        """
+        Get categorical item values by certain item ids in dense format.
+
+        Parameters
+        ----------
+        items: torch.Tensor
+            Internal item ids.
+
+        Returns
+        -------
+        torch.Tensor
+            categorical item values in dense format.
+        """
         # TODO: Add the whole `feature_dense` to the right gpu device at once?
         feature_dense = self.item_features.take(items.detach().cpu().numpy()).get_dense()
         return torch.from_numpy(feature_dense).to(self.device)
 
     @classmethod
     def from_dataset(cls, dataset: Dataset, n_factors: int, dropout_rate: float) -> tpe.Self:
-        """TODO"""
+        """
+        Create CatFeaturesItemNet from RecTools dataset.
+
+        Parameters
+        ----------
+        dataset: Dataset
+            RecTools dataset.
+        n_factors: int
+            Latent embedding size of item embeddings.
+        dropout_rate: float
+            Probability of a hidden unit of item embedding to be zeroed.
+        """
         item_features = dataset.item_features
 
         if item_features is None:
@@ -160,7 +203,7 @@ class IdEmbeddingsItemNet(ItemNetBase):
 
     @property
     def device(self) -> torch.device:
-        """TODO"""
+        """Return device the item net is located."""
         return self.ids_emb.weight.device
 
     @classmethod
@@ -172,9 +215,14 @@ class IdEmbeddingsItemNet(ItemNetBase):
 
 class ItemNetConstructor(ItemNetBase):
     """
-    Base class constructor for ItemNet, taking as input a sequence of ItemNetBase nets,
-    including custom ItemNetBase nets.
-    Constructs item's embedding based on aggregation of its embeddings from the passed networks.
+    Constructed network for item embeddings based on aggregation of embeddings from transferred item network types.
+
+    Parameters
+    ----------
+    n_items: int
+        Number of items in the dataset.
+    item_net_blocks: Sequence(ItemNetBase)
+        Latent embedding size of item embeddings.
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


rectools/models/sasrec.py
```diff
@@ -126,6 +169,11 @@ def from_dataset(cls, dataset: Dataset, n_factors: int, dropout_rate: float) ->
             raise ValueError("`item_features` in `dataset` must be `SparseFeatures` instance.")
 
         item_cat_features = item_features.get_cat_features()
+
+        if item_cat_features.values.size == 0:
```
blondered, line 173: I don't think it's a good idea. We will have this block enabled by default. If dataset doesn't contain categorical features then this embeddings will not be created. But when these features are present they will be used.
This behaviour is the same as for all of our models that work with features (like ALS and Lightfm)

tests/models/test_sasrec.py
```diff
@@ -582,3 +593,349 @@ def test_get_dataloader_recommend(
...
+            dataset_item_features, n_factors=n_factors, dropout_rate=0.5
+        )
+
+        actual_item_features = cat_item_embeddings.item_features
+        actual_n_items = cat_item_embeddings.n_items
+        actual_n_cat_features = cat_item_embeddings.n_cat_features
+        actual_embedding_dim = cat_item_embeddings.category_embeddings.embedding_dim
+
+        expected_item_features = dataset_item_features.item_features
+        # TODO: remove after adding Dense Features support
```
blondered, line 729: Dense features do not support categories. So there will be no support for CatFeaturesItemNet if features are dense.
@feldlime please correct me if I'm wrong

tests/models/test_sasrec.py
```diff
@@ -582,3 +593,349 @@ def test_get_dataloader_recommend(
...
+            np.random.choice(dataset_item_features.item_id_map.internal_ids, size=n_items, replace=False)
+        )
+        cat_item_embeddings = IdEmbeddingsItemNet.from_dataset(
+            dataset_item_features, n_factors=n_factors, dropout_rate=0.5
+        )
+
+        expected_item_ids = cat_item_embeddings(items)
+        assert expected_item_ids.shape == (n_items, n_factors)
+
+    def test_raises_when_dataset_no_features(self) -> None:
```
blondered, line 764: This is not an expected behaviour (see other comment). This block (CatFeaturesItemNet) should be ignored if there are:
1. No features in dataset
2. Features are dense
3. No category features are present in sparse features
This block should just be dropped from the item net.

These would be good cases for tests. Model shouldn't raise errors in any of these cases.

blondered (changes requested):

rectools/models/sasrec.py
```diff
@@ -45,7 +45,7 @@ def get_all_embeddings(self) -> torch.Tensor:
 
     @property
     def device(self) -> torch.device:
```
blondered, line 47: I think we should drop all `device` properties. And stop doing `to(device)` and do `to(<tensor>)` instead. Otherwise we can catch problems with lighting multi-device training. 
Let's change it to lightning advices for device agnostic code https://pytorch-lightning.readthedocs.io/en/1.0.8/multi_gpu.html

We can do it in this PR or in the next one

tests/models/test_sasrec.py
```diff
@@ -582,3 +593,349 @@ def test_get_dataloader_recommend(
...
+
+        actual_n_items = item_id_embeddings.n_items
+        actual_embedding_dim = item_id_embeddings.ids_emb.embedding_dim
+
+        assert actual_n_items == DATASET.item_id_map.size
+        assert actual_embedding_dim == n_factors
+
+    @pytest.mark.parametrize(
+        "n_items,n_factors",
+        (
```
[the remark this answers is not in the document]
In48semenov: Fixed

tests/models/test_sasrec.py
In48semenov: Fixed

rectools/models/sasrec.py
In48semenov: Fixed

tests/models/test_sasrec.py
```diff
@@ -582,3 +593,337 @@ def test_get_dataloader_recommend(
...
+                    [12, "f3", 1],
+                    [13, "f3", 2],
+                    [14, "f3", 3],
+                    [15, "f3", 4],
+                    [17, "f3", 5],
+                    [16, "f3", 6],
+                ],
+                columns=["id", "feature", "value"],
+            )
+            ds = Dataset.construct(
```
blondered, line 885: This is still a dataset with sparse features. To make them dense pass `make_dense_item_features=True`
Also for dense features expected_n_item_net_blocks will be different when `CatFeaturesItemNet` is present

tests/models/test_sasrec.py
```diff
@@ -582,3 +593,337 @@ def test_get_dataloader_recommend(
...
+
+        item_net: ItemNetConstructor = ItemNetConstructor.from_dataset(
+            ds, n_factors=10, dropout_rate=0.5, item_net_block_types=item_net_block_types
+        )
+
+        actual_n_items = item_net.n_items
+        actual_item_net_blocks = len(item_net.item_net_blocks)
+
+        assert actual_n_items == dataset_item_features.item_id_map.size
+        assert actual_item_net_blocks == expected_n_item_net_blocks
```
blondered, line 900:
```suggestion
        assert actual_n_item_net_blocks == expected_n_item_net_blocks
```

tests/models/test_sasrec.py
```diff
@@ -582,3 +593,337 @@ def test_get_dataloader_recommend(
...
+    @pytest.mark.parametrize(
+        "item_net_block_types,sparse_features,expected_n_item_net_blocks",
+        (
+            ((IdEmbeddingsItemNet,), True, 1),
+            ((IdEmbeddingsItemNet, CatFeaturesItemNet), True, 2),
+            ((IdEmbeddingsItemNet,), False, 1),
+            ((IdEmbeddingsItemNet, CatFeaturesItemNet), False, 1),
+        ),
+    )
+    def test_create_from_dataset(
```
blondered, line 865: Please name the test according to what you are testing. Here you test that number of item net block is correct. Let's name it `test_correct_number_of_item_net_blocks`

tests/models/test_sasrec.py
```diff
@@ -582,3 +593,337 @@ def test_get_dataloader_recommend(
...
+            item_features_df=item_features,
+            cat_item_features=["f1", "f2"],
+        )
+        return ds
+
+    def test_catalogue(self) -> None:
+        item_net = ItemNetConstructor.from_dataset(
+            DATASET, n_factors=10, dropout_rate=0.5, item_net_block_types=(IdEmbeddingsItemNet,)
+        )
+        expected_feature_catalogue = torch.arange(0, item_net.n_items)
```
blondered, line 838: You are just reusing exactly the same code that is in the class method. This doesn't make any sense for a test. You can either check it from number of items in dataset. Or just drop this test)

tests/models/test_sasrec.py
```diff
@@ -582,3 +593,337 @@ def test_get_dataloader_recommend(
...
+        )
+        item_net: ItemNetConstructor = ItemNetConstructor.from_dataset(
+            dataset_item_features, n_factors=n_factors, dropout_rate=0.5, item_net_block_types=item_net_block_types
+        )
+
+        expected_embeddings = item_net(items)
+
+        assert expected_embeddings.shape == (n_items, n_factors)
+
+    def test_raise_when_no_item_net_blocks(self) -> None:
```
blondered, line 927: Let's also add a test where we pass only `CatFeaturesItemNet` block and dataset doesn't have item features or has only dense features or has sparse features without categories.

rectools/models/sasrec.py
```diff
@@ -87,45 +90,87 @@ def __init__(
...
-            explanation = """When `use_cat_features_embs` is True, the dataset must have item features."""
-            raise ValueError(explanation)
+            explanation = """When `CatFeaturesItemNet` is used, the dataset must have item features."""
+            warnings.warn(explanation)
+            return None
 
         if not isinstance(item_features, SparseFeatures):
-            raise ValueError("`item_features` in `dataset` must be `SparseFeatures` instance.")
+            explanation = (
+                """When `CatFeaturesItemNet` is used, `item_features` in `dataset` must be `SparseFeatures` instance."""
```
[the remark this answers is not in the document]
In48semenov: Fixed

rectools/models/sasrec.py
```diff
@@ -87,45 +90,87 @@ def __init__(
...
+            )
+            warnings.warn(explanation)
+            return None
 
         item_cat_features = item_features.get_cat_features()
+
+        if item_cat_features.values.size == 0:
+            explanation = """When `CatFeaturesItemNet` is used, the dataset must have item category features."""
```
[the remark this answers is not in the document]
In48semenov: Fixed

tests/models/test_sasrec.py
```diff
@@ -582,3 +593,337 @@ def test_get_dataloader_recommend(
...
+            DATASET, n_factors=10, dropout_rate=0.5, item_net_block_types=(IdEmbeddingsItemNet,)
+        )
+        expected_feature_catalogue = torch.arange(0, item_net.n_items)
+        assert torch.equal(item_net.catalogue, expected_feature_catalogue)
+
+    @pytest.mark.parametrize(
+        "item_net_block_types,n_factors",
+        (
+            ((IdEmbeddingsItemNet,), 8),
+            ((IdEmbeddingsItemNet, CatFeaturesItemNet), 16),
```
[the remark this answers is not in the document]
In48semenov: Fixed

tests/models/test_sasrec.py
In48semenov: Fixed

tests/models/test_sasrec.py
In48semenov: Fixed

tests/models/test_sasrec.py
In48semenov: Fixed

tests/models/test_sasrec.py
In48semenov: Fixed

tests/models/test_sasrec.py
In48semenov: Fixed

tests/models/test_sasrec.py
```diff
@@ -885,6 +883,7 @@ def test_create_from_dataset(
             ds = Dataset.construct(
                 INTERACTIONS,
                 item_features_df=item_features,
+                make_dense_user_features=make_dense_user_features,
```
blondered, line 886: item

tests/models/test_sasrec.py
blondered: You did `make_dense_user_features` instead of **item**

PR #209, opened 2024-11-13
Feature/two_stage

blondered:
`CandidateRankingModel` and tutorial

TODO: doctrsings, tests


poetry.lock
```diff
@@ -293,6 +293,52 @@ d = ["aiohttp (>=3.7.4)", "aiohttp (>=3.7.4,!=3.9.0)"]
...
+    {file = "catboost-1.2.7-cp38-cp38-manylinux2014_x86_64.whl", hash = "sha256:78d2211fb38c31d0ba749eeebc846490c5a298b5f065035fce158c2c8ed0588e"},
+    {file = "catboost-1.2.7-cp38-cp38-win_amd64.whl", hash = "sha256:a1683ac7cdef337bd3490e4aaec11d6fdfee478174bdf7de76a513efa16a1584"},
+    {file = "catboost-1.2.7-cp39-cp39-macosx_11_0_universal2.whl", hash = "sha256:04a0c51ef72741360c90ee037e14466393e487eb1b4f96a95b847524f26be02f"},
+    {file = "catboost-1.2.7-cp39-cp39-manylinux2014_aarch64.whl", hash = "sha256:d2b6aa5f8a41be6f40ae127eedea83450b670788340cac30e74cffb25607c3ba"},
+    {file = "catboost-1.2.7-cp39-cp39-manylinux2014_x86_64.whl", hash = "sha256:e58cf8966e33931acebffbc744cf640e8abd08d0fdfb0e503c107552cea6c643"},
+    {file = "catboost-1.2.7-cp39-cp39-win_amd64.whl", hash = "sha256:90405d3962dd6d0b0960db35dcba10bdea9add112812f011d03043b927f4760e"},
+    {file = "catboost-1.2.7.tar.gz", hash = "sha256:3ed1658bd22c250a12f9c55cf238d654d7a87d9b45f063ec39965a8884a7e9d3"},
+]
+
+[package.dependencies]
```
feldlime, line 330: I know I said we can put catboost easily to dependencies, but I didn't know it brought so many dependencies.

graphviz, matplotlib and plotly are not really needed for rectools. And restriction numpy<2.0 isn't also very good, we're planning to update it

So, I'm thinking it may be a good idea to put catboost to extras. wdyt?

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+from rectools import Columns
+from rectools.dataset import Dataset
+from rectools.dataset.identifiers import ExternalIds
+from rectools.model_selection import Splitter
+from rectools.models.base import ErrorBehaviour, ModelBase, NotFittedError
+
+
+@tp.runtime_checkable
+class ClassifierBase(tp.Protocol):
+    def fit(self, *args: tp.Any, **kwargs: tp.Any) -> None: ...
```
feldlime, line 21: 1. Models usually return themselves after fit
2. Let's put x and y as explicit arguments, otherwise we can't check ourselves. Same for other methods

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+        else:
+            raise ValueError("Got unexpected model_type")
+        reco.sort_values(by=[Columns.User, Columns.Score], ascending=False, inplace=True)
+        return reco
+
+
+class CatBoostReranker(Reranker):
+    def __init__(
+        self,
+        model: tp.Union[CatBoostClassifier, CatBoostRanker] = CatBoostRanker(verbose=False),
```
feldlime, line 79: It's weird linters are silent here.  
But setting a mutable object as a default value is a really bad idea

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+            fit_kwargs.update(self.fit_kwargs)
+
+        return fit_kwargs
+
+    def fit(self, candidates_with_target: pd.DataFrame) -> None:
+        fit_kwargs = self.prepare_fit_kwargs(candidates_with_target)
+        self.model.fit(**fit_kwargs)
+
+    def rerank(self, candidates: pd.DataFrame) -> pd.DataFrame:
+        reco = candidates[Columns.UserItem].copy()
```
feldlime, line 60: image

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+        # TODO: think about it
+        # x_full = x_full[self.model.feature_names_]
+
+        if isinstance(self.model, ClassifierBase):
+            reco[Columns.Score] = self.model.predict_proba(x_full)[:, 1]
+        elif isinstance(self.model, RankerBase):
+            reco[Columns.Score] = self.model.predict(x_full)
+        else:
+            raise ValueError("Got unexpected model_type")
+        reco.sort_values(by=[Columns.User, Columns.Score], ascending=False, inplace=True)
```
feldlime, line 72: I think it'd be good to keep the order of users. At least we keep it in other models afaik

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+        self.pool_kwargs = pool_kwargs
+
+    def prepare_fit_kwargs(self, candidates_with_target: pd.DataFrame) -> tp.Dict[str, tp.Any]:
+        if self.is_classifier:
+            candidates_with_target = candidates_with_target.drop(columns=Columns.UserItem)
+            pool_kwargs = {
+                "data": candidates_with_target.drop(columns=Columns.Target),
+                "label": candidates_with_target[Columns.Target],
+            }
+        elif isinstance(self.model, RankerBase):
```
feldlime, line 95: Looks weird: first check is done by internal flag, but here instead of `else` you check the type again, and check not for `CatBoostRanker`, but for `RankerBase`

We have 3 options:
1. We rely on the type annotations and assume we got the correct model type (one of two possible). Then we create a flag and use it everywhere.
2. We suppose there may be mistakes and check the type every type.
3. There is an intermediate option: check type once in the `__init__` and make a flag. But user can always do `reranker.model = invalid_model`

We need to decide. I'd choose the 3rd one: one the one hand we check the input to keep user from making mistakes, but if they know what they do, then let them do it

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+        super().__init__(model)
+        self.is_classifier = isinstance(model, CatBoostClassifier)
+        self.fit_kwargs = fit_kwargs
+        self.pool_kwargs = pool_kwargs
+
+    def prepare_fit_kwargs(self, candidates_with_target: pd.DataFrame) -> tp.Dict[str, tp.Any]:
+        if self.is_classifier:
+            candidates_with_target = candidates_with_target.drop(columns=Columns.UserItem)
+            pool_kwargs = {
+                "data": candidates_with_target.drop(columns=Columns.Target),
```
feldlime, line 92: Every `drop` creates a copy, and we want to use the memory carefully. Here is no need to make 2 drops, right?

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+                "label": candidates_with_target[Columns.Target],
+            }
+        elif isinstance(self.model, RankerBase):
+            candidates_with_target = candidates_with_target.sort_values(by=[Columns.User])
+            group_ids = candidates_with_target[Columns.User].values
+            candidates_with_target = candidates_with_target.drop(columns=Columns.UserItem)
+            pool_kwargs = {
+                "data": candidates_with_target.drop(columns=Columns.Target),
```
feldlime, line 100: same here

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+        """
+
+        user_features = self._get_user_features(useritem[Columns.User].unique(), dataset, fold_info)
+        item_features = self._get_item_features(useritem[Columns.Item].unique(), dataset, fold_info)
+        useritem_features = self._get_user_item_features(useritem, dataset, fold_info)
+
+        res = useritem
+        res = res.merge(user_features, on=Columns.User, how="left")
+        res = res.merge(item_features, on=Columns.Item, how="left")
+        res = res.merge(useritem_features, on=Columns.UserItem, how="left")
```
feldlime, line 172: It's better to use a chain of operations, otherwise to this moment you will have 4 datasets instead of 1, all are in your memory

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+
+@attr.s(auto_attribs=True)
+class PerUserNegativeSampler(NegativeSamplerBase):
+    num_neg_samples: int = 3
+    random_state: tp.Optional[int] = None
+
+    def sample_negatives(self, train: pd.DataFrame) -> pd.DataFrame:
+        # train: user_id, item_id, scores, ranks, target(1/0)
+
+        negative_mask = train["target"] == 0
```
feldlime, line 190: Columns.Target?

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+        pos = train[~negative_mask]
+        neg = train[negative_mask]
+
+        # Some users might not have enough negatives for sampling
+        num_negatives = neg.groupby([Columns.User])[Columns.Item].count()
+        sampling_mask = train[Columns.User].isin(num_negatives[num_negatives > self.num_neg_samples].index)
+
+        neg_for_sample = train[sampling_mask & negative_mask]
+        neg = neg_for_sample.groupby([Columns.User], sort=False).apply(
```
feldlime, line 199: I see it's experimental, but I don't see TODO, so I'll just remind that `apply` is super super slow and we're not using it.  
I'm ok with TODO comment here if you don't want fixing it now

feldlime: A possible solution (maybe not the best one, but without using apply)

1. After calculating `num_negatives` calculate also the sampling ratio as `minimum(self.num_neg_samples / num_negatives, 1)`
2. Convert it to series with user_id as index and map it to the `neg` dataset
3. Add a column with random values in [0, 1)
4. Filter out rows with random values > sampling ratio

You also don't need a sampling mask here.  
And you can avoid splitting to pos and neg parts simply putting `1` as a sampling ratio for positives. This also means you don't need to do shuffling in the end. And I really don't like this shuffling, because:
1. It's a heavy operation
2. You have `sample` method in the `Sampler` class and it's quite not obvious that you shuffle the data inside

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+
+@attr.s(auto_attribs=True)
+class NegativeSamplerBase:
+    def sample_negatives(self, train: pd.DataFrame) -> pd.DataFrame:
+        raise NotImplementedError()
+
+
+@attr.s(auto_attribs=True)
+class PerUserNegativeSampler(NegativeSamplerBase):
+    num_neg_samples: int = 3
```
feldlime, line 184: `n_neg_samples`?

tests/models/test_candidate_ranking.py
```diff
@@ -0,0 +1,259 @@
...
+        sampled_df = sampler.sample_negatives(sample_data)
+
+        # Check if the resulting DataFrame has the correct columns
+        assert set(sampled_df.columns) == set(sample_data.columns)
+
+        # Check if the number of negatives per user is correct
+        for user_id in sampled_df[Columns.User].unique():
+            user_data = sampled_df[sampled_df[Columns.User] == user_id]
+            num_negatives = len(user_data[user_data[Columns.Target] == 0])
+            assert num_negatives == num_neg_samples
```
feldlime, line 45:
```suggestion
    n_negatives_per_user = data.groupby(Columns.User)["Target"].agg(lambda target: (target == 0).sum())
    assert (n_negatives_per_user == num_neg_samples).all()
```

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+        users: ExternalIds,
+        dataset: Dataset,
+        filter_viewed: bool,
+        for_train: bool,
+        items_to_recommend: tp.Optional[ExternalIds] = None,
+        on_unsupported_targets: ErrorBehaviour = "raise",
+    ) -> pd.DataFrame:
+
+        if for_train and not self.is_fitted_for_train:
+            raise NotFittedError(self.model.__class__.__name__)
```
feldlime, line 304: Let's add that this model isn't fitted for train/recommend

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+        -------
+        tp.Dict[str, CandidateGenerator]
+            Dictionary with candidate generator identifiers as keys and candidate generators as values.
+        """
+        model_count = {}
+        cand_gen_dict = {}
+        for candgen in candidate_generators:
+            model_name = candgen.model.__class__.__name__
+            if model_name not in model_count:
+                model_count[model_name] = 0
```
feldlime, line 387:
```suggestion
        model_counts = defaultdict(int)
        cand_gen_dict = {}
        for candgen in candidate_generators:
            model_name = candgen.model.__class__.__name__
```

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+            model_count[model_name] += 1
+            identifier = f"{model_name}_{model_count[model_name]}"
```
feldlime, line 389: I don't yet know how exactly do you plan to use identifiers, but do you think it make sense to generate some sort of hash for models?
Or, if they are used for column names (=features), maybe it's better to ask user to provide the names? In this case it would be easier to explore feature importance e.g.

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+                on_unsupported_targets=on_unsupported_targets,
+            )
+
+            # Process ranks and scores as features
+            rank_col_name, score_col_name = f"{identifier}_rank", f"{identifier}_score"
+
+            candidates.rename(
+                columns={Columns.Rank: rank_col_name, Columns.Score: score_col_name},
+                inplace=True,
+                errors="ignore",
```
feldlime, line 559: Please add comment here why we ignore errors

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+        try:
+            candidates = self._get_candidates_from_first_stage(
+                users=users,
+                dataset=dataset,
+                filter_viewed=filter_viewed,
+                items_to_recommend=items_to_recommend,
+                for_train=False,
+                on_unsupported_targets=on_unsupported_targets,
+            )
+        except NotFittedError:
```
feldlime, line 618: Why do we consider 2 possible valid scenarios here (models are fitted and not fitted)?

rectools/models/candidate_ranking.py
```diff
@@ -0,0 +1,637 @@
...
+            self._fit_candidate_generators(dataset, for_train=False)
+            candidates = self._get_candidates_from_first_stage(
+                users=users,
+                dataset=dataset,
+                filter_viewed=filter_viewed,
+                items_to_recommend=items_to_recommend,
+                for_train=False,
+                on_unsupported_targets=on_unsupported_targets,
+            )
```
feldlime, line 627: Anyway let's avoid code duplication

feldlime: smth like  
```
if not all(canidate.is_fitted in self.candidates):
    self._fit...
candidates = self._get...
```

rectools/models/candidate_ranking.py
blondered: We need a dict for feature names in columns, yes.
But user (me) is too lazy to create a dict when he can just pass a list.
It's just a bit faster. I would prefer to leave it at it is.

rectools/models/candidate_ranking.py
blondered: In real service it is common that we use a fitted reranker but want to fit first stage on fresh data for inference.
I also have a todo comment here that "we should probably add an argument to force fitting again on new dataset"
Maybe we also should pass an argument to `fit` if we want to skip fitting first-stage on full dataset after reranker was fitted. Again - this model can be used in real services where it is not needed to fit first-stage on full dataset after reranker was fitted.

poetry.lock
blondered: I agree that it's better in extras. Fixed

rectools/models/candidate_ranking.py
blondered: 1. Fixed
2. I don't see hot to put X and y in fit since they may have very diffent types. For example Catboost can accept Pool in X and have y as None.

rectools/models/candidate_ranking.py
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: Fixed. But I did the first variant. We do it in many cases in framework. No need for extra checking types when user passes strange things.

rectools/models/candidate_ranking.py
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: Fixed n_negatives

tests/models/test_candidate_ranking.py
blondered: Fixed

tests/models/test_candidate_ranking.py
```diff
@@ -0,0 +1,259 @@
...
+        sample_data.loc[sample_data[Columns.User] == 1, Columns.Target] = [1, 0, 1, 0]
+
+        sampler = PerUserNegativeSampler(num_neg_samples=3, random_state=42)
+        sampled_df = sampler.sample_negatives(sample_data)
+
+        # Check if the resulting DataFrame has the correct columns
+        assert set(sampled_df.columns) == set(sample_data.columns)
+
+        # Check if the number of negatives per user is correct
+        for user_id in sampled_df[Columns.User].unique():
```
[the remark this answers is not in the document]
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: I don't remember why it's here :)
Removed

rectools/models/candidate_ranking.py
blondered: Added flags to both fit and recommend. Gonna add tests in next PR

rectools/models/candidate_ranking.py
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: Added TODO

tests/models/test_candidate_ranking.py
```diff
@@ -0,0 +1,259 @@
...
+
+    @pytest.fixture
+    def model(self) -> PopularModel:
+        return PopularModel()
+
+    def test_not_fitted_errors(self, dataset: Dataset, model: PopularModel, users: tp.List[int]) -> None:
+        generator = CandidateGenerator(model, 2, False, False)
+
+        with pytest.raises(NotFittedError):
+            generator.generate_candidates(users, dataset, filter_viewed=True, for_train=True)
```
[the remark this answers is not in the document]
blondered: Fixed

tests/models/test_candidate_ranking.py
```diff
@@ -0,0 +1,259 @@
...
+        with pytest.raises(NotFittedError):
+            generator.generate_candidates(users, dataset, filter_viewed=True, for_train=False)
+
+        generator.fit(dataset, for_train=True)
+
+        generator.generate_candidates(users, dataset, filter_viewed=True, for_train=True)
```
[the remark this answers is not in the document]
blondered: Fixed

rectools/models/candidate_ranking.py
blondered: Fixed

rectools/models/candidate_ranking.py
feldlime: Add there a link to this conversation pls

rectools/models/candidate_ranking.py
blondered: Can't see a way to make it faster. This implementation is still the fastest out of alternatives.

feldlime: First, it worth to note that in your current solution user order is not preserving, you need to add `sort=False` in the `groupby` method. Also please add a test on this.

Second, I'm getting warning (pandas version is `2.2.0`)
image

feldlime: I tried couple of solutions, they may be faster with medium number of candidates per user (<100) but are slow with bigger numbers.

However I've checked polars and it's really good here. I think it may make sense to add it since sorting in pandas is really slow. wdyt?
image

rectools/models/ranking/candidate_ranking.py
```diff
@@ -0,0 +1,567 @@
...
+            filter_viewed=filter_viewed,
+            items_to_recommend=items_to_recommend,
+            for_train=False,
+            on_unsupported_targets=on_unsupported_targets,
+        )
+
+        train = self.feature_collector.collect_features(candidates, dataset, fold_info=None)
+
+        reco = self.reranker.rerank(train)
+        reco = reco.groupby(Columns.User).head(k)
```
feldlime, line 562: `sort = False`

Add a test on it please

rectools/models/candidate_ranking.py
feldlime: And one more important thing: we very likely can speed this up independently of chosen method using `partition` algorithm. Instead of sorting 300 candidates and selecting then top-10, it's much faster to pick 10 largest scores and then sort them, as we did before the 'ImplicitRanker` for vector models.

blondered:
Pandas version was merged https://github.com/MobileTeleSystems/RecTools/pull/234

PR #211, opened 2024-11-26
kwargs in recommend

spirinamayya:
Added kwargs in recommend of ModelBase. Changed signature for PopularModel
Added dtype change

### DIFF
--- a/rectools/models/base.py
+++ b/rectools/models/base.py
@@ -96,6 +96,7 @@ class ModelBase:
         items_to_recommend: tp.Optional[ExternalIds] = None,
         add_rank_col: bool = True,
         on_unsupported_targets: ErrorBehaviour = "raise",
+        **kwargs: tp.Any,
     ) -> pd.DataFrame:
         r"""
         Recommend items for users.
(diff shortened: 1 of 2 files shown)


rectools/models/base.py
```diff
@@ -398,7 +398,8 @@ def recommend(
         """
         self._check_is_fitted()
         self._check_k(k)
-
+        # `dataset.item_id_map.external_dtype` can change
+        original_item_type = dataset.item_id_map.external_dtype
```
blondered, line 402: Let's do the same logic for users. Let's assume that users dtype could also be broken during `_custom_transform_dataset_u2i`

spirinamayya: fixed

blondered (approved):

rectools/models/base.py
```diff
@@ -398,7 +398,9 @@ def recommend(
         """
         self._check_is_fitted()
         self._check_k(k)
-
+        # `dataset.user_id_map.external_dtype`, `dataset.item_id_map.external_dtype` can change
```
feldlime, line 401: Why can they change?

It's not allowed to change function param

feldlime (changes requested):

rectools/models/base.py
blondered: We are not changing dataset itself. But we do call `dataset = self._custom_transform_dataset_u2i(dataset, users, on_unsupported_targets)`
and here we get new object under `dataset` name.

rectools/models/base.py
```diff
@@ -398,7 +398,9 @@ def recommend(
...
+        original_user_type = dataset.user_id_map.external_dtype
```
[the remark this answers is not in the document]
spirinamayya: fixed

rectools/models/base.py
spirinamayya: fixed

feldlime (approved):

PR #212, opened 2024-11-27
BERT4Rec tutorial

spirinamayya:
Added BERT4Rec tutorial


blondered:
" Item embeddings from these sequences are fed to multi-head self-attention to acquire user sequence latent representation"
That's not exactly true. Embeddings are fed into transformer blocks which have more login inside then just multi-head self-attention

blondered:
"After one or several stacked attention blocks, resulting embeddings are used to predict the next item."
Ambiguous.
resulting embeddings are used to predict all items in the sequence. Each item is being predicted based only on previous items information.

blondered:
About main differences in the table. Let's go further.
1. You didn't specify losses. Original implementations of these models had different losses. 
2. Some of the differences are conceptual (target and attention direction). They cannot be excluded from model. But other (losses, activation functions, transformer blocks) could be a subject of change. Try to explicitly emphasize it. It really is important.

blondered:
Let's add to RecTools implementation section references for our implementations (which implementations we took as goals, what we changed). Let'a also emphasize that we included number of negatives as a hyper-perameter.

blondered:
Let's remove `def recommend` from `basic usage section`.
We merge recos with items only once. No need for a separate method

blondered:
In losses section please add you description of what's happening. What is the difference between losses?

blondered:
Let's remove `TODO: Add category item features embeddings`
We will add them afterwards

blondered:
Transformer models -> Transformer models tutorial

blondered:
Categorical feature embeddings are dealt with separately from id embeddings and are concatenated after weight update.

1. They are not concatenated. They are summed up.
2. Why "after weight update"?

You can just write "Categorical feature embeddings are summed up with other embeddings for each item if they are present in the model."

PR #213, opened 2024-11-30
Feature/load model

feldlime:
## Description

- Added `load_model` function

## Type of change

- [ ] Optimization

## How Has This Been Tested?

### DIFF
--- a/tests/models/test_serialization.py
+++ b/tests/models/test_serialization.py
@@ -1,0 +1,41 @@
+import typing as tp
+from tempfile import NamedTemporaryFile
+
+import pytest
+from implicit.als import AlternatingLeastSquares
+from implicit.nearest_neighbours import ItemItemRecommender
+from lightfm import LightFM
+
+from rectools.models import (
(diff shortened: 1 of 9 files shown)


tests/models/test_serialization.py
```diff
@@ -0,0 +1,51 @@
...
+    load_model,
+)
+from rectools.models.base import ModelBase
+
+from .utils import get_final_successors
+
+MODEL_CLASSES = [
+    cls
+    for cls in get_final_successors(ModelBase)
+    if cls.__module__.startswith("rectools.models") and not (sys.version_info >= (3, 12) and cls is LightFMWrapperModel)
```
blondered, line 29: What's wring with LightFM and 3.12?

feldlime: they are incompatible with each other (we also skip all lightfm tests for 3.12)

blondered (approved):

blondered (approved):

PR #215, opened 2024-12-03
Feature/transformers loss

blondered:
Moved losses away from lightning module


rectools/models/sasrec.py
```diff
@@ -1131,17 +1215,17 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-locals
         use_key_padding_mask: bool = False,
         dropout_rate: float = 0.2,
         session_max_len: int = 32,
-        loss: str = "softmax",
+        loss: Union[Literal["softmax", "BCE", "gBCE"], LossCalculatorBase] = "softmax",
```
feldlime, line 1218: Let's put `Literal["softmax", "BCE", "gBCE"]` into a separate var at the top

blondered: fixed

rectools/models/sasrec.py
```diff
@@ -628,20 +630,20 @@ class SequenceDataset(TorchDataset):
...
-    sessions: List[List[int]]
+    sessions: tp.List[tp.List[int]]
         User sessions in the form of sequences of items ids.
-    weights: List[List[float]]
+    weights: tp.List[tp.List[float]]
         Weight of each interaction from the session.
     """
 
-    def __init__(self, sessions: List[List[int]], weights: List[List[float]]):
+    def __init__(self, sessions: tp.List[tp.List[int]], weights: tp.List[tp.List[float]]):
```
feldlime, line 639: I think we can add some aliases like `Session = tp.List[int]`, `SessionWeights = tp.List[float]`, etc.  
Should be easier to read

rectools/models/sasrec.py
```diff
@@ -932,125 +934,71 @@ def _collate_fn_train(
...
-
-    def _get_pos_neg_logits(self, x: torch.Tensor, y: torch.Tensor, negatives: torch.Tensor) -> torch.Tensor:
-        # [n_items + n_item_extra_tokens, n_factors], [batch_size, session_max_len, n_factors]
-        item_embs, session_embs = self.torch_model(x)
-        pos_neg = torch.cat([y.unsqueeze(-1), negatives], dim=-1)  # [batch_size, session_max_len, n_negatives + 1]
-        pos_neg_embs = item_embs[pos_neg]  # [batch_size, session_max_len, n_negatives + 1, n_factors]
-        # [batch_size, session_max_len, n_negatives + 1]
-        logits = (pos_neg_embs @ session_embs.unsqueeze(-1)).squeeze(-1)
-        return logits
+    @property
```
feldlime, line 982: Why is it a property? Does it depend on a specific instance? For me it looks like a class attribute

feldlime: If I'm right, please also use `tp.ClassVar` in the type annotation (just found it recently)

rectools/models/sasrec.py
```diff
@@ -21,6 +21,8 @@
 from rectools.models.rank import Distance, ImplicitRanker
 from rectools.types import InternalIdsArray
 
+LossDescription = tp.Literal["softmax", "BCE", "gBCE"]
```
feldlime, line 24: It's not a description at all  
Maybe `LossName`?

rectools/models/sasrec.py
```diff
@@ -1069,41 +1017,199 @@ def _calc_softmax_loss(cls, logits: torch.Tensor, y: torch.Tensor, w: torch.Tens
...
+    n_actual_items: int
+        Number of actual items in catalog. Needed for calculating negatives sampling rate for gBCE loss.
+    """
+
+    def __init__(
+        self,
+        torch_model: TransformerBasedSessionEncoder,
+        loss: tp.Union[LossDescription, SessionEncoderHeadBase],
+        lr: float,
+        gbce_t: float,
```
feldlime, line 1147: Should be optional, right? Since it's used only for one specific loss

rectools/models/sasrec.py
```diff
@@ -1179,6 +1288,15 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-locals
         self.loss = loss
         self.gbce_t = gbce_t
 
+    @property
+    def requires_negatives(self) -> bool:
+        """Indicator whether negatives are needed for loss computation during training."""
+        if isinstance(self.loss, SessionEncoderHeadBase):
+            return self.loss.requires_negatives
+        if self.loss == "softmax":
```
feldlime, line 1296: You have this property in every class, why do you need to have this logic here again?

rectools/models/sasrec.py
blondered: It' more convenient to have actual default params for such things. It this case we have out-of-the-box decisions for all loss variants. n_negatives is also needed for 2 losses out of 3

rectools/models/sasrec.py
blondered: I wouldn't do that. We have sessions as Lists of Lists of ints and we have Sessions as torch.Tensors. So here I would leave it as it is now. It is more clear which data we are using right now.

rectools/models/sasrec.py
blondered: we don't have SessionEncoderHead initialised from LossName yet. so we can't say if it requires negatives from initialised Head.
But we need to know this for data_preaparator initialisation. And data preparator is needed before constructing torch nets because torch nets depend on data_preaparator results.

rectools/models/sasrec.py
blondered: fixed

rectools/models/sasrec.py
feldlime: But here are no default values...

Also n_negatives I see used only for 1 loss

rectools/models/sasrec.py
```diff
@@ -1071,41 +1018,191 @@ def _calc_softmax_loss(cls, logits: torch.Tensor, y: torch.Tensor, w: torch.Tens
...
-        Encode user sessions.
-        """
+    def predict_step(self, batch: tp.Dict[str, torch.Tensor], batch_idx: int) -> torch.Tensor:
+        """Prediction step. Encode user sessions. Get only last position hidden state from each user session."""
         encoded_sessions = self.torch_model.encode_sessions(batch["x"], self.item_embs.to(self.device))[:, -1, :]
         return encoded_sessions
 
+    def configure_optimizers(self) -> torch.optim.Adam:
+        """Choose what optimizers and learning-rate schedulers to use in optimization"""
+        adam_betas = (0.9, 0.98)
```
feldlime, line 1202: Is it going to be a hardcoded value?

feldlime: If you plan to change it later, please add TODO here

rectools/models/sasrec.py
```diff
@@ -934,125 +936,70 @@ def _collate_fn_train(
...
+class DotProductSoftmaxHead(SessionEncoderHeadBase):
+    """
+    Applies so called "Softmax" Loss function which stands for Cross Entropy Loss
+    which is calculated over full items catalog.
+    Logits are computed as dot product of session embeddings and item embeddings.
+    """
 
-        pos_logits = logits[:, :, 0:1].to(torch.float64)
-        neg_logits = logits[:, :, 1:].to(torch.float64)
+    requires_negatives: tp.ClassVar[bool] = False
```
feldlime, line 994: you don't need to replicate the annotation every time  
Actually I don't know the recommendations here. But we don't do it in other places

rectools/models/sasrec.py
feldlime: Sorry the code is too complicated for me to deep dive into it now. The only thing I know that having duplicated logic is a really bad idea. Let's please don't do this

rectools/models/sasrec.py
blondered: We have default value  for bot `gbce_t` and `n_negatives` in model initialisation. So we always assume it's a float and an int.

n_negatives is used for BCE and gBCE, not used for softmax

rectools/models/sasrec.py
blondered: We really can't not do this.
We do not initialise torch nets until we get dataset (and Head is a torch net).
We can't get dataset until we get data_preaparator.
We can't get data_preparator until we know if need negatives or not.

Many things depend in whether we need negatives or not.

rectools/models/sasrec.py
blondered: It's gonna be hardcoded. Can't move all of the params to user, We have too many small things here and there.

rectools/models/sasrec.py
blondered: We can initialise heads during model initialisation just to get `require_negatives` value. And save received loss value for `get_config`. But I'm not sure it's better this way.

blondered:
Decided not to do this feature for user convenience

PR #218, opened 2024-12-04
Feature/vector rank params

### LINKED ISSUE #199: recommend() method for ImplicitALSWrapperModel

blondered:
`recommend_n_threads` and `recommend_use_gpu_ranking` arguments to EASEModel and vector models
Closes #199


rectools/models/ease.py
```diff
@@ -65,21 +72,33 @@ class EASEModel(ModelBase[EASEModelConfig]):
     def __init__(
         self,
         regularization: float = 500.0,
-        num_threads: int = 1,
+        num_threads: int = 1,  # TODO: decide. We already have it. But this is actually recommend_cpu_n_threads
```
feldlime, line 75: I think it should be `recommend_cpu_n_threads` (`recommend_n_threads`) in the end. 

So, let's do next:
1. Add new param `recommend_n_threads` with correct default value (btw, why do we use 1, not 0?)
2. Change here default value to `None`
3. If this param is not `None` we will use it value, but raise a deprecation warning
4. If this param is `None` we will use `recommend_n_threads` value

rectools/models/ease.py
```diff
@@ -54,7 +56,12 @@ class EASEModel(ModelBase[EASEModelConfig]):
     verbose : int, default 0
         Degree of verbose output. If 0, no output will be provided.
     num_threads: int, default 1
-        Number of threads used for `recommend` method.
+        Number of threads used for recommendation ranking on cpu.
+    recommend_use_gpu_ranking: Optional[bool], default ``None``
+        Flag to use gpu for recommendation ranking. If ``None``, `implicit.gpu.HAS_CUDA` will be
+        checked before inference.
+        This attribute can be changed manually before calling model `recommend` method if you
```
feldlime, line 63: Don't mix up method arguments and class attributes. They are not necessarily named the same. You should specify here which exactly attribute to change.

rectools/models/ease.py
```diff
@@ -93,6 +112,13 @@ def _fit(self, dataset: Dataset) -> None:  # type: ignore
         self.weight = np.array(gram_matrix_inv / (-np.diag(gram_matrix_inv)))
         np.fill_diagonal(self.weight, 0.0)
 
+    @property
+    def _recommend_use_gpu_ranking(self) -> bool:
+        use_gpu = HAS_CUDA
+        if self.recommend_use_gpu_ranking is False:
+            use_gpu = False
```
feldlime, line 119:
```suggestion
        use_gpu = self.recommend_use_gpu_ranking is not False and HAS_CUDA
```

rectools/models/implicit_als.py
```diff
@@ -133,22 +146,41 @@ class ImplicitALSWrapperModel(VectorModel[ImplicitALSWrapperModelConfig]):
...
+    def __init__(
+        self,
+        model: AnyAlternatingLeastSquares,
+        verbose: int = 0,
+        fit_features_together: bool = False,
+        recommend_cpu_n_threads: tp.Optional[int] = None,
+        recommend_use_gpu_ranking: tp.Optional[bool] = None,
+    ):
+        self._config = self._make_config(
+            model, verbose, fit_features_together, recommend_cpu_n_threads, recommend_use_gpu_ranking
```
feldlime, line 158: let's use named assignments here to avoid mistakes

rectools/models/ease.py
```diff
@@ -65,21 +72,33 @@ class EASEModel(ModelBase[EASEModelConfig]):
     def __init__(
         self,
         regularization: float = 500.0,
-        num_threads: int = 1,
+        num_threads: int = 1,  # TODO: decide. We already have it. But this is actually recommend_cpu_n_threads
         verbose: int = 0,
+        recommend_use_gpu_ranking: tp.Optional[bool] = None,
```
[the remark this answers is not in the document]
blondered: fixed

rectools/models/ease.py
blondered: fixed only for EASE before we approve final words

rectools/models/ease.py
blondered: Fixed. Not sure what to do with tests and configs.
I removed `num_threads` from config. And didn't do any tests for it.

rectools/models/ease.py
blondered: fixed and removed from property because we call it only in one method

rectools/models/ease.py
```diff
@@ -51,10 +54,22 @@ class EASEModel(ModelBase[EASEModelConfig]):
     ----------
     regularization : float
         The regularization factor of the weights.
+    num_threads: Optional[int], default ``None``
+        Deprecated. Number of threads used for recommendation ranking on cpu.
+    recommend_n_threads: int, default 0
+        Number of threads to use for recommendation ranking on cpu.
+        If you want to change ranking behaviour, you can manually assign new value to model
+        `recommend_n_threads` attribute. This should be done before calling  model
+        `recommend` method.
```
feldlime, line 63:
```suggestion
         If you want to change this parameter after model is initialized, 
         you can manually assign new value to model `recommend_n_threads` attribute.
```

Wdyt? If agreed, lets's change for both params

rectools/models/ease.py
```diff
@@ -51,10 +54,22 @@ class EASEModel(ModelBase[EASEModelConfig]):
     ----------
     regularization : float
         The regularization factor of the weights.
+    num_threads: Optional[int], default ``None``
+        Deprecated. Number of threads used for recommendation ranking on cpu.
```
feldlime, line 58:
```suggestion
        Deprecated, use `recommend_n_threads` instead. 
        Number of threads used for recommendation ranking on cpu.
```

rectools/models/ease.py
```diff
@@ -65,23 +80,45 @@ class EASEModel(ModelBase[EASEModelConfig]):
...
         super().__init__(verbose=verbose)
         self.weight: np.ndarray
         self.regularization = regularization
-        self.num_threads = num_threads
+
+        if num_threads is not None:
+            warnings.warn(
+                """
+            `num_threads` argument is deprecated and will be removed in future releases.
+            Please use `recommend_n_threads` instead")
```
feldlime, line 97:
```suggestion
            Please use `recommend_n_threads` instead".
```

rectools/models/ease.py
```diff
@@ -117,7 +154,8 @@ def _recommend_u2i(
             k=k,
             filter_pairs_csr=ui_csr_for_filter,
             sorted_object_whitelist=sorted_item_ids_to_recommend,
-            num_threads=self.num_threads,
+            num_threads=self.recommend_n_threads,
+            use_gpu=self.recommend_use_gpu_ranking is not False and HAS_CUDA,
```
feldlime, line 158:
```suggestion
            use_gpu=self.recommend_use_gpu_ranking and HAS_CUDA,
```

rectools/models/ease.py
blondered: agree. fixed

rectools/models/ease.py
blondered: fixed

rectools/models/ease.py
blondered: fixed

rectools/models/ease.py
blondered: fixed

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `ImplicitBPRWrapperModel` model ([#232](https://github.com/MobileTeleSystems/RecTools/pull/232))
+- All vector models and `EASEModel` support for enabling ranking on gpu and selecting number of threads for cpu ranking. Added `recommend_n_threads` and `recommend_use_gpu_ranking` parameters to `EASEModel`, `ImplicitALSWrapperModel`, `ImplicitBPRWrapperModel`, `PureSVDModel` and `DSSMModel`. Added `recommend_use_gpu_ranking` to `LightFMWrapperModel`. Gpu and cpu ranking may provide different ordering of items with identical scores in recommendation table, so this could change ordering items in recommendations since gpu ranking is now used as a default one. ([#218](https://github.com/MobileTeleSystems/RecTools/pull/218))
```
feldlime, line 12:
```suggestion
- All vector models and `EASEModel` support for enabling ranking on GPU and selecting the number of threads for CPU ranking. Added `recommend_n_threads` and `recommend_use_gpu_ranking` parameters to `EASEModel`, `ImplicitALSWrapperModel`, `ImplicitBPRWrapperModel`, `PureSVDModel` and `DSSMModel`. Added `recommend_use_gpu_ranking` to `LightFMWrapperModel`. GPU and CPU ranking may provide different ordering of items with identical scores, so this could change recommendations since GPU ranking is now used as a default one. ([#218](https://github.com/MobileTeleSystems/RecTools/pull/218))
```

rectools/models/lightfm.py
```diff
@@ -105,7 +107,21 @@ class LightFMWrapperModel(FixedColdRecoModelMixin, VectorModel[LightFMWrapperMod
     epochs: int, default 1
         Will be used as `epochs` parameter for `LightFM.fit`.
     num_threads: int, default 1
-        Will be used as `num_threads` parameter for `LightFM.fit`.
+        Will be used as `num_threads` parameter for `LightFM.fit`. Should be larger then 0.
+        This will also be used as number of threads to use for recommendation ranking on CPU.
+        If you want to change number of threads for ranking after model is initialized,
+        you can manually assign new value to model `recommend_n_threads` attribute.
```
feldlime, line 113:
```suggestion
        Can also be used as number of threads for recommendation ranking on CPU.
        See `recommend_n_threads` for details.
```

rectools/models/lightfm.py
```diff
@@ -131,6 +149,13 @@ def __init__(
         self._model = model
         self.n_epochs = epochs
         self.n_threads = num_threads
+        self._recommend_n_threads = recommend_n_threads
```
feldlime, line 152:
```suggestion
        self._recommend_n_threads = recommend_n_threads  # used to make a config 
```

rectools/models/lightfm.py
```diff
@@ -131,6 +149,13 @@ def __init__(
...
+        self.recommend_n_threads = 0
+        if recommend_n_threads is not None:
+            self.recommend_n_threads = recommend_n_threads
+        elif num_threads > 0:
```
feldlime, line 156: By the docstring it's always >0, so we don't need this condition here

rectools/models/rank.py
```diff
@@ -149,18 +151,23 @@ def _rank_on_gpu(
         object_norms: tp.Optional[np.ndarray],
         filter_query_items: tp.Optional[tp.Union[sparse.csr_matrix, sparse.csr_array]],
     ) -> tp.Tuple[np.ndarray, np.ndarray]:  # pragma: no cover
-        object_factors = implicit.gpu.Matrix(object_factors.astype(np.float32))
+
+        def _convert_arr_to_implicit_gpu_matrix(arr: np.ndarray) -> implicit.gpu.Matrix:
+            # We need to explicitly create copy to handle transposed arrays correctly
```
feldlime, line 156:
```suggestion
            # We need to explicitly create copy to handle transposed arrays correctly
            # since Matrix is created from a direct copy of the underlying memory block, and `.T` is just a view
```

rectools/models/rank.py
```diff
@@ -149,18 +151,23 @@ def _rank_on_gpu(
         object_norms: tp.Optional[np.ndarray],
         filter_query_items: tp.Optional[tp.Union[sparse.csr_matrix, sparse.csr_array]],
     ) -> tp.Tuple[np.ndarray, np.ndarray]:  # pragma: no cover
-        object_factors = implicit.gpu.Matrix(object_factors.astype(np.float32))
+
+        def _convert_arr_to_implicit_gpu_matrix(arr: np.ndarray) -> implicit.gpu.Matrix:
```
feldlime, line 155: We also create `gpu.Matrix` in the ALSWrapper code

Let's please make this function an util and use everywhere

rectools/models/rank.py
```diff
@@ -252,6 +259,7 @@ def rank(  # pylint: disable=too-many-branches
                 filter_query_items=filter_query_items,
             )
         else:
+            check_blas_config()
```
feldlime, line 262: Please add a comment why do we need this

rectools/models/vector.py
```diff
@@ -40,7 +41,8 @@ class VectorModel(ModelBase[ModelConfig_T]):
 
     u2i_dist: Distance = NotImplemented
     i2i_dist: Distance = NotImplemented
-    n_threads: int = 0  # TODO: decide how to pass it correctly for all models
+    recommend_n_threads: int = 0
```
feldlime, line 44: Since we accept these params in every vector model in init, why do we still need them here?

blondered: Because we are accessing it in `VectorModel` methods?

CHANGELOG.md
blondered: fixed

rectools/models/lightfm.py
blondered: fixed

rectools/models/lightfm.py
blondered: fixed

rectools/models/lightfm.py
blondered: fixed

rectools/models/rank.py
blondered: fixed

rectools/models/rank.py
blondered: fixed

rectools/models/rank.py
blondered: I was pretty sure that it affects implicit ranking speed as well as it affects implicit training speed. But now I've checked again and it's not the case. Removed.

rectools/models/rank.py
blondered: I remembered now why I didn't do this as a public function. When implicit is built without cuda, it cannot import it because of the return type annotation. Wo we really need it?

feldlime: No, I'm not saying we need this public. I just want to add this logic with copy wherever we create gpu.Matrix

blondered: fixed

rectools/models/vector.py
blondered: fixed

tests/models/test_implicit_bpr.py
```diff
@@ -241,11 +282,18 @@ def test_i2i(
             actual,
         )
 
-    @pytest.mark.skip("BPR doesn't behave deterministically")
     def test_second_fit_refits_model(self, dataset: Dataset, use_gpu: bool) -> None:
-        base_model = BayesianPersonalizedRanking(factors=8, num_threads=2, use_gpu=use_gpu, random_state=1)
+        # note that num_threads > 1 will make model training undeterministic
+        # https://github.com/benfred/implicit/issues/710
+        # GPU training is always undeterministic so we only test for CPU training
+        base_model = BayesianPersonalizedRanking(factors=8, num_threads=1, use_gpu=False, random_state=1)
```
feldlime, line 289:
```suggestion
        # GPU training is always nondeterministic so we only test for CPU training
        if use_gpu:
            pytest.skip("BPR is nondeterministic on GPU")
        base_model = BayesianPersonalizedRanking(factors=8, num_threads=1, use_gpu=use_gpu, random_state=1)
```
(To avoid having 2 identical tests)

feldlime (approved):

PR #230, opened 2024-12-11
SASRec gpu/cpu multi-device tests

spirinamayya:
Added gpu/cpu multi-device tests


tests/models/test_sasrec.py
```diff
@@ -93,33 +162,49 @@ def trainer(self) -> Trainer:
...
+                    {
+                        Columns.User: [10, 10, 10, 30, 30, 30, 40, 40, 40],
+                        Columns.Item: [13, 14, 15, 14, 13, 12, 12, 17, 14],
+                        Columns.Rank: [1, 2, 3, 1, 2, 3, 1, 2, 3],
+                    }
+                ),
             ),
         ),
     )
     # TODO: tests do not pass for multiple GPUs
```
blondered, line 182: remove

tests/models/test_sasrec.py
```diff
@@ -24,7 +24,34 @@
 
 from .data import DATASET, INTERACTIONS
 
+# pylint: disable=too-many-lines
 
+
+@pytest.mark.parametrize(
```
blondered, line 30: Let's not do this for all of the tests for this model. We are going to write more of them.
Let's do it only for test_u2i.

tests/models/test_sasrec.py
spirinamayya: fixed

tests/models/test_sasrec.py
spirinamayya: fixed

blondered (approved):

Issue #132, opened 2024-05-14
implicit BPR-MF wrapper model


blondered:
### Feature Description

Create a wrapper for popular BPR-MF baseline

### Why this feature?

It's easy to implement and it's often used as a baseline in research

chezou:
Created an experimental PR. https://github.com/MobileTeleSystems/RecTools/pull/232

I know there are several discussion points, but I hope it clarifies the desired specs.

blondered:
@chezou thank you so much for your contributions! I've merged PR with BPR model.

I wouldn't focus on adding features for BPR since it's a complicated task and will take a lot of time.
BPR itself is one of the most known baselines. But since we already have an MF algorithm with features we don't necessarily need another one

Right now in RecTools we are focusing on adding features and providing maximum customisation to transformer models (SASRec and BERT4Rec). They can be found here: https://github.com/MobileTeleSystems/RecTools/blob/experimental/sasrec/examples/tutorials/transformers_tutorial.ipynb
And we are preparing them for the release.

Another big story right now is CandidateRankingModel which uses baseline models to generate candidates and then uses Gradient Boosting to rerank them. https://github.com/MobileTeleSystems/RecTools/blob/experimental/two_stage/examples/tutorials/candidate_ranking_model_tutorial.ipynb

As for the baselines, we still need SLIM model. It is well known both for quality and efficiency. And I do have a feeling that we are missing it in the framework right now. We also have an issue for it: https://github.com/MobileTeleSystems/RecTools/issues/103
We have found SLIM to be quite useful in production for fast MVPs and candidate generation, but feature selection (as in the paper) is a must otherwise it's learning too slow.
Inference for SLIM could be done with `implicit` framework ItemKNN inference code which is optimised already. And SLIM has sparse item-item similarity matrix under the hood so inference logic is the same.

chezou:
Thanks for the review and comment. I don't have any preference for having the feature for BPR, so feel free to close this issue.

I briefly researched SLIM (specifically, SLIM ElasticNet) and found that many frameworks use this CPU-based implementation: https://github.com/MaurizioFD/RecSys2019_DeepLearning_Evaluation/blob/master/SLIM_ElasticNet/SLIMElasticNetRecommender.py, or some use a PyTorch-based one.

Anyway, let's continue to discuss on #103

PR #227, opened 2024-12-10
Support Python 3.13

chezou:
## Description
Support Python 3.13, including:

- [x] torch
- [x] cupy
- [x] lightfm
- [x] pytorch_lightning

excluding:

- nmslib https://github.com/metabrainz/nmslib-metabrainz/pull/5
  - nmslib-metabrainz only supports until Python 3.12.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Optimization

## How Has This Been Tested?
Before submitting a PR, please check yourself against the following list. It would save us quite a lot of time.
- Have you read the contribution guide?
- Have you updated the relevant docstrings? We're using Numpy format, please double-check yourself
- Does your change require any new tests?
- Have you updated the changelog file?

PR #232, opened 2024-12-11
[Experimental] Introduce ImplicitBPRWrapper model

chezou:
## Description

Address [#132](https://github.com/MobileTeleSystems/RecTools/issues/132)

As a first step, don't support features at this moment since it requires to additional design for CPU implementation because CPU BPR class is fully Cythonized.

While I tried to keep deterministic for unit tests, BPR updates random states every fit, so I gave up some tests. Some tests may flakey.

## Type of change

- [ ] Optimization

## How Has This Been Tested?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,6 +125,9 @@ line-length = 120
 target-version = ["py39", "py310", "py311", "py312"]
 
 
+[tool.isort]
+profile = "black"
+
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,9 +125,6 @@ line-length = 120
 target-version = ["py39", "py310", "py311", "py312"]
 
 
-[tool.isort]
-profile = "black"
-
 [build-system]
 requires = ["poetry-core"]
 build-backend = "poetry.core.masonry.api"
--- a/rectools/metrics/intersection.py
+++ b/rectools/metrics/intersection.py
@@ -12,7 +12,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from typing import Dict, Hashable, Optional, Union
+from collections.abc import Hashable
+from typing import Dict, Optional, Union
 
 import attr
 import numpy as np
--- a/tests/models/test_implicit_bpr.py
+++ b/tests/models/test_implicit_bpr.py
@@ -1,4 +1,5 @@
 import typing as tp
+from copy import deepcopy
 
 import implicit.gpu
 import numpy as np
@@ -19,6 +20,7 @@ from implicit.gpu.bpr import (
 from rectools.columns import Columns
 from rectools.dataset.dataset import Dataset
 from rectools.exceptions import NotFittedError
+from rectools.models.base import ModelBase
 from rectools.models.implicit_bpr import (
     AnyBayesianPersonalizedRanking,
     ImplicitBPRWrapperModel,
@@ -28,7 +30,6 @@ from tests.models.data import DATASET
 from tests.models.utils import (
     assert_default_config_and_default_model_params_are_the_same,
     assert_dumps_loads_do_not_change_model,
-    assert_get_config_and_from_config_compatibility,
     assert_second_fit_refits_model,
 )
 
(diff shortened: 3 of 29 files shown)


pyproject.toml
```diff
@@ -125,6 +125,9 @@ line-length = 120
 target-version = ["py39", "py310", "py311", "py312"]
 
 
+[tool.isort]
+profile = "black"
```
feldlime, line 129: While `pyproject.toml` is considered as a preferred format of configuration for `isort`, we continue using `setup.cfg` instead for it (and for most of other tools as well) since we think this way it's easier to separate external project characteristics (description, supported language versions, dependencies, etc.) and development tools settings.  
So, please move this to the `setup.cfg`

feldlime: P.S. Having `tool.isort` section here makes `isort` ignore settings in the `setup.cfg`, that's why you have reformatted imports in many project files

chezou: The reason why I changed the isort format is I faced formatting conflicts around import and pylint comment.

I'm okay to move it to setup.cfg, but many tools has been migrated to pyproject.toml already as much as possible.

This should be a separate topic though.

chezou: Moved the configuration into setup.cfg  5f3211d

tests/models/test_implicit_bpr.py
```diff
@@ -0,0 +1,435 @@
...
+                True,
+                {10: {13, 17}, 20: {17}},
+            ),
+            (
+                False,
+                {10: {11, 13, 17}, 20: {11, 13, 17}},
+            ),
+        ),
+    )
+    def test_with_allowlist(
```
blondered, line 159: Could you please rename this to `test_with_whitelist` to keep consistency within the framework?

rectools/metrics/intersection.py
```diff
@@ -12,7 +12,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from typing import Dict, Hashable, Optional, Union
+from collections.abc import Hashable
```
blondered, line 15: Why do we need this change?

blondered:
Hi @chezou! Thank you so much for this PR, this is very profound. I couldn't find any fixes for non deterministic behaviour. So just did a few small comments.
@feldlime is also looking at this.

tests/models/test_implicit_bpr.py
chezou: Okay.

I was trying to follow the [ACM guideline](https://www.acm.org/diversity-inclusion/words-matter#:~:text=friendly%20force%2C%20etc.-,blacklist/whitelist,-%3A%20terms%20used%20to). Using these terminologies could cause flagging on ethic concerns in the paper review such [as ICLR](https://openreview.net/forum?id=Z1MKx8mNu9&noteId=F5JZsdpYBN).

rectools/metrics/intersection.py
chezou: This is because `typing.Hashable` will be deprecated, and Python 3.12 warns it https://github.com/python/cpython/issues/94309
It causes lint error on Python 3.12

chezou:
I found the RTD error:

```
poetry export -f requirements.txt -o requirements.txt -E all --without-hashes
The command "export" does not exist.
```

This is because Poetry no longer supports `export` command itself and need to install plugin.
https://python-poetry.org/docs/cli/#export

However, poetry export is not stable enough, and I suggested to change on https://github.com/MobileTeleSystems/RecTools/pull/227#discussion_r1879089413

tests/models/test_implicit_bpr.py
feldlime: Thank you for this comment. I agree, it's an important topic and we should using correct terminologies. I think we should create a separate PR, mark `whitelist` as deprecated everywhere, and add `allowlist`.

chezou: Thanks for your suggestion. I committed to making it into a whitelist for now.
[69c92f4](https://github.com/MobileTeleSystems/RecTools/pull/232/commits/69c92f48990bd09eaaa70848dc3dd23629f1a2ab)

feldlime (approved):

feldlime (approved):
Thanks for the PR @chezou, looks great!

chezou:
Thanks for the review!

Note that I tried to stabilize the test as much as possible [e70ef91](https://github.com/MobileTeleSystems/RecTools/pull/232/commits/e70ef91dbca92a45ad550eee883c68e598c60dcf), however, I see some flakiness presumably due to lack of holding initial random state like rectools-lightfm.
If we face flaky test issues, let's skip them.

PR #238, opened 2025-01-10
Added validation example for SASRec model.

In48semenov:
## Description

 Validation example for SASRec model.

## Type of change

- [ ] Optimization

## How Has This Been Tested?


rectools/models/sasrec.py
```diff
@@ -730,9 +732,81 @@ def n_item_extra_tokens(self) -> int:
...
-    def process_dataset_train(self, dataset: Dataset) -> Dataset:
+    def _prepared_dataset_train_for_val(self, dataset: Dataset, raw_interactions: pd.DataFrame) -> Dataset:
         """TODO"""
-        interactions = dataset.get_raw_interactions()
+        interactions_train = raw_interactions.copy()
+        dataset_train = dataset
+        if self.val_k_out:
+            # Ctreating train dataset
+            interactions_train[f"{Columns.Rank}_inverse"] = (
+                interactions_train.sort_values(Columns.Datetime, ascending=False).groupby(Columns.User).cumcount() + 1
```
feldlime, line 742: Please specify explicitly `sort` in groupby

rectools/models/sasrec.py
```diff
@@ -730,9 +732,81 @@ def n_item_extra_tokens(self) -> int:
...
+            )
+            mask_train = ~(
+                (interactions_train[Columns.User].isin(self.val_users))
+                & (interactions_train[f"{Columns.Rank}_inverse"].isin(range(1, self.val_k_out + 1)))
```
feldlime, line 746: It's more clear and efficient to use  `<= self.val_k_out` here

rectools/models/sasrec.py
```diff
@@ -730,9 +732,81 @@ def n_item_extra_tokens(self) -> int:
...
 
-    def process_dataset_train(self, dataset: Dataset) -> Dataset:
+    def _prepared_dataset_train_for_val(self, dataset: Dataset, raw_interactions: pd.DataFrame) -> Dataset:
         """TODO"""
-        interactions = dataset.get_raw_interactions()
+        interactions_train = raw_interactions.copy()
+        dataset_train = dataset
+        if self.val_k_out:
+            # Ctreating train dataset
+            interactions_train[f"{Columns.Rank}_inverse"] = (
```
feldlime, line 741: Please make this ranks as a separate pd.Series, not a column in the dataframe.

You still can use it in the filter, but you don't need to add it to the dataframe and then drop

rectools/models/sasrec.py
```diff
@@ -730,9 +732,81 @@ def n_item_extra_tokens(self) -> int:
...
+            dataset_train = Dataset(user_id_map, dataset.item_id_map, interactions_train, item_features=item_features)
+
+        return dataset_train
+
+    def _processed_interactions_val(self, dataset_train: Dataset, raw_interactions: pd.DataFrame) -> pd.DataFrame:
+        """TODO"""
+        interactions_val = raw_interactions.copy()
+        interactions_val = interactions_val[
+            (interactions_val[Columns.User].isin(self.val_users))
+            & (interactions_val[Columns.User].isin(dataset_train.user_id_map.to_external))
```
feldlime, line 770: 1. Instead of `to_external` use `external_ids`, it's faster and code is clearer
2. Instead of doing 2 conditions for the User column, just combine the 2 sets of users you need here and make just one condition

rectools/models/sasrec.py
```diff
@@ -730,9 +732,81 @@ def n_item_extra_tokens(self) -> int:
...
+            & (interactions_val[Columns.Item].isin(dataset_train.item_id_map.to_external))
+        ]
+        interactions_val[f"{Columns.Rank}_inverse"] = (
```
feldlime, line 773: Same here about groupby, separate pd.Series and <=

Seems here we have a duplicate in the code, can we avoid this?

rectools/models/sasrec.py
```diff
@@ -730,9 +732,81 @@ def n_item_extra_tokens(self) -> int:
...
+            interactions_val.sort_values(Columns.Datetime, ascending=False).groupby(Columns.User).cumcount() + 1
+        )
+        mask_val = interactions_val[f"{Columns.Rank}_inverse"].isin(range(1, self.val_k_out + 1))
+        interactions_val.drop(columns=f"{Columns.Rank}_inverse", inplace=True)
+
+        interactions_val.loc[~mask_val, Columns.Weight] = 0
+        interactions_val = (
+            interactions_val.sort_values(Columns.Datetime)
+            .groupby(Columns.User)
```
feldlime, line 782: please put `sort` in groupby everywhere (`sort=False` if you don't need groups to be sorted, True otherwise)

rectools/models/sasrec.py
```diff
@@ -905,6 +1019,26 @@ def _collate_fn_recommend(
 class SASRecDataPreparator(SessionEncoderDataPreparatorBase):
     """TODO"""
 
+    def __init__(
+        self,
+        session_max_len: int,
+        batch_size: int,
+        dataloader_num_workers: int,
+        shuffle_train: bool = True,
+        item_extra_tokens: tp.Sequence[tp.Hashable] = (PADDING_VALUE,),
```
feldlime, line 1028: Since python 3.12  Hashable should be imported from `collections.abc` instead of `typing`

examples/tutorials/validate_transformers_turorial.ipynb
```diff
@@ -0,0 +1,2310 @@
+{
```
blondered, line 1: Do we need to save `self.prev_interactions` for `ValidationMetrics` callback? I don't see why we need it

rectools/models/sasrec.py
```diff
@@ -730,15 +732,29 @@ def n_item_extra_tokens(self) -> int:
         """Return number of padding elements"""
         return len(self.item_extra_tokens)
 
-    def process_dataset_train(self, dataset: Dataset) -> Dataset:
+    def process_dataset_train(
+        self,
+        dataset: Dataset,
+        split_interactions_train_val: tp.Optional[tp.Callable] = None,
```
blondered, line 738: About the namings. I suggest:
split_interactions_train_val -> get_val_mask_func
self.processed_dataset_train -> self.train_dataset
self.processed_interactions_val -> self.val_interactions
val_max_k -> top_k_saved_val_reco

if we will move metrics computation to "on_validation_batch_end" callback, then we don't need "val_max_k" at all (which is nice but not obligatory) (but let's keep it for now until everything else is finalized)

rectools/models/sasrec.py
```diff
@@ -1007,17 +1152,53 @@ def training_step(self, batch: Dict[str, torch.Tensor], batch_idx: int) -> torch
...
+        else:
+            raise ValueError(f"loss {self.loss} is not supported")
+
+        self.log("train/loss", train_loss, on_step=False, on_epoch=True, prog_bar=self.verbose > 0)
+        return train_loss
+
+    def validation_step(self, batch: Dict[str, torch.Tensor], batch_idx: int) -> torch.Tensor:
+        """Validate step."""
+        x, y, w = batch["x"], batch["y"], batch["yw"]
+        all_last_logits = None
```
blondered, line 1173:
```suggestion
        logits = None
```

rectools/models/sasrec.py
```diff
@@ -1007,17 +1152,53 @@ def training_step(self, batch: Dict[str, torch.Tensor], batch_idx: int) -> torch
...
+        if self.loss == "softmax":
+            logits = self._get_full_catalog_logits(x)
+            last_logits = logits[:, -1:, :]
+            val_loss = self._calc_softmax_loss(last_logits, y, w)
+            all_last_logits = logits[:, -1, :]
```
blondered, line 1178:
```suggestion
            logits = self._get_full_catalog_logits(x)[:, -1:, :]
            val_loss = self._calc_softmax_loss(logits, y, w)
```

rectools/models/sasrec.py
```diff
@@ -1007,17 +1152,53 @@ def training_step(self, batch: Dict[str, torch.Tensor], batch_idx: int) -> torch
...
+        elif self.loss == "BCE":
             negatives = batch["negatives"]
             logits = self._get_pos_neg_logits(x, y, negatives)
```
blondered, line 1181:
```suggestion
            pos_neg_logits = self._get_pos_neg_logits(x, y, negatives)[:, -1:, :]
```

rectools/models/sasrec.py
```diff
@@ -1007,17 +1152,53 @@ def training_step(self, batch: Dict[str, torch.Tensor], batch_idx: int) -> torch
...
-            return self._calc_bce_loss(logits, y, w)
-        if self.loss == "gBCE":
+            last_logits = logits[:, -1:, :]
+            val_loss = self._calc_bce_loss(last_logits, y, w)
```
blondered, line 1183:
```suggestion
            val_loss = self._calc_bce_loss(pos_neg_logits, y, w)
```

rectools/models/sasrec.py
```diff
@@ -1007,17 +1152,53 @@ def training_step(self, batch: Dict[str, torch.Tensor], batch_idx: int) -> torch
...
-            return self._calc_gbce_loss(logits, y, w, negatives)
+            last_logits = logits[:, -1:, :]
+            val_loss = self._calc_gbce_loss(last_logits, y, w, negatives)
+        else:
+            raise ValueError(f"loss {self.loss} is not supported")
+
+        self.log("val/loss", val_loss, on_step=False, on_epoch=True, prog_bar=self.verbose > 0)
+
+        if self.val_max_k is not None:
+            if all_last_logits is None:
```
blondered, line 1195:
```suggestion
            if logits is None:
```

examples/tutorials/validate_transformers_turorial.ipynb
```diff
@@ -0,0 +1,2189 @@
...
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
```
blondered, line 5: `epoch_val_recos` and `epoch_targets` can also be obtained with `itertools.chain.from_iterable`

examples/tutorials/validate_transformers_turorial.ipynb
In48semenov: Removed from `self` variable.

examples/tutorials/validate_transformers_turorial.ipynb
```diff
@@ -0,0 +1,2189 @@
+{
+ "cells": [
+  {
```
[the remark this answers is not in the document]
In48semenov: **done**

examples/tutorials/validate_transformers_turorial.ipynb
In48semenov: **done**

rectools/models/sasrec.py
In48semenov: **done**

rectools/models/sasrec.py
In48semenov: **done**

rectools/models/sasrec.py
In48semenov: **done**

rectools/models/sasrec.py
In48semenov: **done**

rectools/models/sasrec.py
In48semenov: **done**

rectools/models/sasrec.py
In48semenov: **done**

rectools/models/sasrec.py
In48semenov: **done**

rectools/models/sasrec.py
feldlime: don't see the change, still all groupby are without`sort` argument

rectools/models/sasrec.py
feldlime: don't see the change, still all groupby are withoutsort argument :(

rectools/models/sasrec.py
```diff
@@ -1007,17 +1150,50 @@ def training_step(self, batch: Dict[str, torch.Tensor], batch_idx: int) -> torch
...
+        elif self.loss == "gBCE":
+            negatives = batch["negatives"]
+            pos_neg_logits = self._get_pos_neg_logits(x, y, negatives)[:, -1:, :]
+            val_loss = self._calc_gbce_loss(pos_neg_logits, y, w, negatives)
+        else:
+            raise ValueError(f"loss {self.loss} is not supported")
+
+        self.log("val/loss", val_loss, on_step=False, on_epoch=True, prog_bar=self.verbose > 0)
+
+        if self.top_k_saved_val_reco is not None:
```
blondered, line 1189: I did some thinking and finally I strongly believe that `on_validation_batch_end` logic with outputs processing in a callback will be a much cleaner solution.
1. It requires less memory for cache. It is important for big dataset with leave-one-out scheme. 
2. It provides total flexibility. User will not have to inherit and rewrite validation step for custom validation logic. He will only need to write callbacks.
3. Recos can be computed with filtering viewed items or any other logic. Popularity-sampled metrics can also be computed with s special callback. And it will be clean.

Let's do the following:
1. We need to do `self.item_embs = self.torch_model.item_model.get_all_embeddings()` in our main code. I think the correct hook is `on_validation_epoch_start`.
2. We shouldn't compute anything apart for the logits that are actually needed for correct loss. Predict step shouldn't think about what will be needed by metrics callbacks. It's not clean to prepare full logits here for losses that don't need them. User may not need them in his custom metrics callbacks.
3. We should prepare a dict for outputs of all the things that were actually computed. For all logits let's use "logits" key. For `pos_neg_logits` logits let's use "pos_neg_logits" key. For missing logits let's put None under the key.
4. We can process results in `on_validation_batch_end` logic in metrics callback. We can compute missing logits there if they are needed and not provided by the outputs.

examples/tutorials/validate_transformers_turorial.ipynb
```diff
@@ -0,0 +1,841 @@
...
+    "**Custom function for split data on train and validation:**"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 13,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def get_val_mask(interactions: pd.DataFrame, val_k_out: int, val_users: ExternalIds):\n",
```
blondered, line 520: We can't show `val_k_out` logic in the tutorial until we handle k_out targets in our losses correctly. Or users will be extremely confused.
Let's drop it from the tutorial. Let's show only leave-one-out everywhere.

We can support these cases later but in the next releases, not the first one.

rectools/models/sasrec.py
```diff
@@ -1184,28 +1361,28 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-locals
...
             gbce_t=self.gbce_t,
-            n_item_extra_tokens=n_item_extra_tokens,
+            data_preparator=self.data_preparator,
+            verbose=self.verbose,
+            top_k_saved_val_reco=self.top_k_saved_val_reco,
         )
 
         self.fit_trainer = deepcopy(self._trainer)
-        self.fit_trainer.fit(self.lightning_model, train_dataloader)
+        self.fit_trainer.fit(self.lightning_model, train_dataloader, val_dataloader)
```
blondered, line 1385: We should clean data_preparator processed data after this

rectools/models/sasrec.py
```diff
@@ -1007,17 +1150,50 @@ def training_step(self, batch: Dict[str, torch.Tensor], batch_idx: int) -> torch
...
+        else:
+            raise ValueError(f"loss {self.loss} is not supported")
 
-        raise ValueError(f"loss {self.loss} is not supported")
+        self.log("train/loss", train_loss, on_step=False, on_epoch=True, prog_bar=self.verbose > 0)
+        return train_loss
+
+    def validation_step(self, batch: Dict[str, torch.Tensor], batch_idx: int) -> torch.Tensor:
+        """Validate step."""
+        x, y, w = batch["x"], batch["y"], batch["yw"]
```
blondered, line 1170: Please write tensor shapes in the comments. They are different from the train step shapes if I'm not mistaken

blondered:
In the next PR: let's return val_mask and drop_mask instead of one mask

rectools/models/sasrec.py
In48semenov: Sorry, now added.

examples/tutorials/validate_transformers_turorial.ipynb
In48semenov: **done**

rectools/models/sasrec.py
In48semenov: we agreed that it would be in the next PR

rectools/models/sasrec.py
In48semenov: **done**

tests/models/nn/test_sasrec.py
```diff
@@ -456,47 +465,34 @@ def test_warn_when_hot_user_has_cold_items_in_recommend(
...
-        self, interactions_df: pd.DataFrame, expected_sessions: List[List[int]], expected_weights: List[List[float]]
-    ) -> None:
-        actual = SequenceDataset.from_interactions(interactions_df)
-        assert len(actual.sessions) == len(expected_sessions)
-        assert all(
-            actual_list == expected_list for actual_list, expected_list in zip(actual.sessions, expected_sessions)
-        )
-        assert len(actual.weights) == len(expected_weights)
-        assert all(actual_list == expected_list for actual_list, expected_list in zip(actual.weights, expected_weights))
+        if model.fit_trainer.logger is not None:
```
blondered, line 489: If we by some reason will not have logger this test will fail ti check functionality. Let's explicitly create trainer with CSVLogger for this test and remove `if`.
To make it work create a temporary directory for the test. Example is here: https://github.com/MobileTeleSystems/RecTools/blob/9c2e62f47bc03d8d4958243c5e607aa157c9c748/tests/models/test_sasrec.py#L474

tests/models/nn/test_sasrec.py
```diff
@@ -456,47 +465,34 @@ def test_warn_when_hot_user_has_cold_items_in_recommend(
...
-                [30, 12, 2, "2021-11-26"],
-                [30, 15, 1, "2021-11-25"],
-                [40, 11, 1, "2021-11-25"],
-                [40, 17, 8, "2021-11-26"],
-                [50, 16, 1, "2021-11-25"],
-                [10, 14, 1, "2021-11-28"],
-            ],
-            columns=Columns.Interactions,
+    @pytest.mark.parametrize("verbose", (0, 1))
+    def test_log_metrics(self, dataset: Dataset, trainer: Trainer, verbose: int) -> None:
```
blondered, line 474: Let's add `parametrize` on get_val_mask_func. This is also going to be a happy path for fitting model with val mask.

tests/models/nn/test_sasrec.py
```diff
@@ -566,11 +588,63 @@ def test_process_dataset_train(
...
+                            [0, 3, 0.0, "2021-11-26"],
+                            [0, 2, 1.0, "2021-11-27"],
+                        ],
+                        columns=[Columns.User, Columns.Item, Columns.Weight, Columns.Datetime],
+                    ),
+                ),
+            ),
+        ),
+    )
+    def test_process_dataset_train_val(
```
blondered, line 631:
```suggestion
    def test_process_dataset_with_val_mask(
```

tests/models/nn/test_sasrec.py
```diff
@@ -660,14 +734,36 @@ def test_tranform_dataset_i2i(
...
+                }
+            ),
+        ),
+    )
+    def test_get_dataloader_val(
+        self, dataset: Dataset, data_preparator_val_mask: SASRecDataPreparator, val_batch: tp.List
+    ) -> None:
+        data_preparator_val_mask.process_dataset_train(dataset)
+        dataloader = data_preparator_val_mask.get_dataloader_val()
+        if dataloader is not None:
```
blondered, line 762: This is bad. You might have `get_dataloader_val` broken and you will not know it from this test. The test will pass.
Please remove `if`

tests/models/nn/test_sasrec.py
```diff
@@ -531,7 +527,33 @@ def dataset(self) -> Dataset:
...
+        val_users = [10, 30]
+
+        def get_val_mask(interactions: pd.DataFrame, val_users: ExternalIds) -> pd.Series:
+            rank = (
+                interactions.sort_values(Columns.Datetime, ascending=False, kind="stable")
+                .groupby(Columns.User, sort=False)
+                .cumcount()
+                + 1
+            )
+            val_mask = (interactions[Columns.User].isin(val_users)) & (rank <= val_k_out)
```
blondered, line 546: just put `1` here explicitly

rectools/models/nn/transformer_base.py
```diff
@@ -0,0 +1,630 @@
...
+            outputs["pos_neg_logits"] = pos_neg_logits.squeeze()
+        elif self.loss == "gBCE":
+            negatives = batch["negatives"]
+            pos_neg_logits = self._get_pos_neg_logits(x, y, negatives)[:, -1:, :]
+            outputs["loss"] = self._calc_gbce_loss(pos_neg_logits, y, w, negatives)
+            outputs["pos_neg_logits"] = pos_neg_logits.squeeze()
+        else:
+            raise ValueError(f"loss {self.loss} is not supported")
+
+        self.log("val/loss", outputs["loss"], on_step=False, on_epoch=True, prog_bar=self.verbose > 0)
```
blondered, line 350: Let's add "val/loss" to class attribute `val_loss_name` just like we do for i2i and ui2 distance in vector models.
Same for "train/loss".

rectools/models/sasrec.py
In48semenov: **done**

rectools/models/__init__.py
```diff
@@ -39,6 +39,8 @@
 from .ease import EASEModel
 from .implicit_als import ImplicitALSWrapperModel
 from .implicit_knn import ImplicitItemKNNWrapperModel
+from .nn.bert4rec import BERT4RecModel
```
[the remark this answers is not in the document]
In48semenov: **done**

tests/models/nn/test_sasrec.py
In48semenov: **done**

tests/models/nn/test_sasrec.py
In48semenov: **done**

tests/models/nn/test_sasrec.py
In48semenov: Removed `if` and added `type: ignore`, because `data_preparator_val_mask.get_dataloader_val()` can return None object, but this case isn't related to this test.

tests/models/nn/test_sasrec.py
In48semenov: **done**

rectools/models/nn/transformer_base.py
In48semenov: I added class attribute `val_loss_name`, `train_loss_name` to `SessionEncoderLightningModuleBase` and we can pass when call model's method fit.

blondered: Let's please move them from `fit` arguments to `TransformerModelBase` class attributes. It's gonna be easier to access by user and a bit cleaner.
And we try to avoid adding anything to `fit` as long as we can :)

rectools/models/nn/sasrec.py
```diff
@@ -68,6 +68,33 @@ def _collate_fn_train(
...
+            # take only first target for leave-one-strategy
+            target_idx = [idx for idx, weight in enumerate(ses_weights) if weight != 0][0]
+
+            # ses: [session_len] -> x[i]: [session_max_len]
+            x[i, -len(input_session) :] = input_session[-self.session_max_len :]
+            y[i, -1:] = ses[target_idx]  # y[i]: [1]
+            yw[i, -1:] = ses_weights[target_idx]  # yw[i]: [1]
+
+        batch_dict = {"x": torch.LongTensor(x), "y": torch.LongTensor(y), "yw": torch.FloatTensor(yw)}
+        # TODO: we are sampling negatives for paddings
```
blondered, line 88: No TODO

rectools/models/nn/transformer_base.py
In48semenov: **done**

rectools/models/nn/sasrec.py
In48semenov: **done**

tests/models/nn/test_sasrec.py
```diff
@@ -470,29 +487,51 @@ def test_warn_when_hot_user_has_cold_items_in_recommend(
...
+        is_val_mask_func: bool,
+        expected_columns: tp.List[str],
+    ) -> None:
+        with TemporaryDirectory() as dirname:
+            trainer = Trainer(
+                max_epochs=2,
+                min_epochs=2,
+                deterministic=True,
+                accelerator="cpu",
+                logger=CSVLogger(save_dir=dirname),
```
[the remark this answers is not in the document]
In48semenov: **done**

blondered (approved):

PR #247, opened 2025-01-21
nn modules tests

spirinamayya:
Added docstrings and tests for nn modules


rectools/models/nn/bert4rec.py
```diff
@@ -82,7 +82,13 @@ def _collate_fn_train(
         self,
         batch: List[Tuple[List[int], List[float]]],
     ) -> Dict[str, torch.Tensor]:
-        """TODO"""
+        """
+        Mask session elements to receive `x`.
+        Get target by replacing session elements with 0 with probability 1 - `mask_prob`.
```
blondered, line 87:
```suggestion
        Get target by replacing session elements with a MASK token with probability `mask_prob`.
```

rectools/models/nn/bert4rec.py
```diff
@@ -82,7 +82,13 @@ def _collate_fn_train(
...
+        Truncate each session and target from right to keep (`session_max_len` + 1) last items.
+        Do left padding until  (session_max_len + 1) is reached.
+        If `n_negatives` is not None, generate negative items from uniform distribution.
```
blondered, line 90: `session_max_len + 1` is used here to keep that logic that `session_max_len` is maximum sessions length that model will process during inference. But one extra "MASK" token will be added for making predictions. Thus we need to see `session_max_len + 1` elements during training.

rectools/models/nn/bert4rec.py
```diff
@@ -82,7 +82,13 @@ def _collate_fn_train(
...
+        """
```
blondered, line 91: Please change negatives logic below to have `size=(batch_size, self.session_max_len + 1, self.n_negatives)`
otherwise there will be tensor shapes mismatch during training with negatives.
Please let's add at least one negatives loss test for BERT4Rec, otherwise we will not check that our logic is correct.

rectools/models/nn/transformer_data_preparator.py
```diff
@@ -213,7 +212,19 @@ def get_dataloader_train(self, processed_dataset: Dataset) -> DataLoader:
         return train_dataloader
 
     def get_dataloader_recommend(self, dataset: Dataset) -> DataLoader:
-        """TODO"""
+        """
+        Construct recommend dataloader from processed dataset.
+
+        Parameters
+        ----------
+        processed_dataset : Dataset
```
blondered, line 220: Can't rename arguments in docstrings

tests/models/nn/test_bert4rec.py
```diff
@@ -0,0 +1,538 @@
+import typing as tp
```
blondered, line 1: To fix missing directory error during tests please set `enable_checkpointing=False` in `TransformerModelBase` default trainer.
Also we will use only 1 device by default (because changing this number is the first thing that we do everywhere). Also cross-validate will be correct to run only on one device.

Final default arguments for default trainer are here: https://github.com/MobileTeleSystems/RecTools/blob/976c14a72b9c71122ed406a9278c259c931e6969/rectools/models/nn/transformer_base.py#L598

Don't mind code logic change in my code. For now just change the arguments.

tests/models/nn/test_transformer_data_preparator.py
```diff
@@ -59,3 +66,186 @@ def test_from_interactions(
...
         assert len(actual.weights) == len(expected_weights)
         assert all(actual_list == expected_list for actual_list, expected_list in zip(actual.weights, expected_weights))
+
+
+class TestSessionEncoderDataPreparatorBase:
+
+    def setup_method(self) -> None:
+        self._seed_everything()
+
+    def _seed_everything(self) -> None:
```
blondered, line 76: I don't think that we need it for data preparator. All is determentistic as far as I see.

rectools/models/nn/bert4rec.py
```diff
@@ -82,7 +82,13 @@ def _collate_fn_train(
         self,
         batch: List[Tuple[List[int], List[float]]],
     ) -> Dict[str, torch.Tensor]:
-        """TODO"""
+        """
```
blondered, line 85: While I was testing configs, I found that we can't do `session_max_len=session_max_len + 1` here https://github.com/MobileTeleSystems/RecTools/blob/5413b1492b620d625a4d69ca5e5d5b62eccfad08/rectools/models/nn/bert4rec.py#L248

This breaks the logic of inheritance and configs. Let's remove +1. Please do this in this PR to check that all tests pass (to avoid changing expected outputs we can just change argument while initializing the model for the test.

tests/models/nn/test_bert4rec.py
```diff
@@ -0,0 +1,538 @@
...
+    def dataset_hot_users_items(self, interactions_df: pd.DataFrame) -> Dataset:
+        return Dataset.construct(interactions_df[:-4])
+
+    @pytest.fixture
+    def trainer(self) -> Trainer:
+        return Trainer(
+            max_epochs=2,
+            min_epochs=2,
+            deterministic=True,
+            accelerator="cpu",
```
blondered, line 60: `enable_checkpointing=False` everywhere in trainer initialization to fix missing directory error

rectools/models/nn/bert4rec.py
spirinamayya: fixed

rectools/models/nn/bert4rec.py
spirinamayya: fixed

rectools/models/nn/transformer_data_preparator.py
spirinamayya: fixed

tests/models/nn/test_bert4rec.py
spirinamayya: fixed

rectools/models/nn/bert4rec.py
```diff
@@ -273,14 +284,14 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-locals
...
-            session_max_len=session_max_len + 1,
+            session_max_len=session_max_len,
             trainer=trainer,
             item_net_block_types=item_net_block_types,
             pos_encoding_type=pos_encoding_type,
             lightning_module_type=lightning_module_type,
         )
         self.data_preparator = data_preparator_type(
-            session_max_len=session_max_len,
+            session_max_len=session_max_len - 1,
```
blondered, line 294: Let's try to clean all of the `+1` and `-1` logic in our data preparators.

The only reason why we pass here `session_max_len - 1` to always do `session_max_len + 1` in Bert data preparator is because we have this `+ 1` here: https://github.com/MobileTeleSystems/RecTools/blob/2a594395bef11d826d69f6ca4c2f76197b783ded/rectools/models/nn/transformer_data_preparator.py#L162

Let's add a class attribute to each data preparator class `train_session_max_len_addition`. base class and bert will have 0. sasrec will have 1.
In base data preparator we will do   `.tail(self.session_max_len + self.train_session_max_len_addition)`

We will also:
- not pass `-1` during bert data preparator initialization
- not add `+1` everywhere in bert data preparator where it is now

This will clean things a lot. And prevent mistakes on custom models that will be created later.

blondered: Let'a also clean all the comments concerning +1 for session_max_len after this is done.

spirinamayya: fixed

blondered (approved):

PR #257, opened 2025-02-04
Feature/transformer tutorial update

blondered:
Transformer draft for final tutorial


blondered:
Please remove unnecessary cells at the beginning of the notebook.

blondered:
Please make image of our architecture smaller. I think we can dot it by adding white space on sides

PR #270, opened 2025-02-27
Experimental/two stage

olesyabulgakova:
We compare differences in branches


blondered:
Difference check not needed any more> merge request will be from https://github.com/MobileTeleSystems/RecTools/tree/experimental/two_stage branch

PR #272, opened 2025-03-18
Feature/transformers fusion cosine

In48semenov:
Added `similarity_module_type` to transformer models and `DistanceSimilarityModule` with `dot` and `cosine` distance options.


blondered:
Decided to move it as a similarity module class
@feldlime this is not ready FYI

rectools/models/nn/transformers/similarity.py
```diff
@@ -0,0 +1,110 @@
...
+        user_ids: InternalIdsArray,
+        k: int,
+        sorted_item_ids_to_recommend: InternalIdsArray,
+        ui_csr_for_filter: tp.Optional[sparse.csr_matrix],
+    ) -> InternalRecoTriplet:
+        """Recommend to users."""
+        raise NotImplementedError()
+
+
+class SimilarityModuleDistance(SimilarityModuleBase):
```
blondered, line 43:
```suggestion
class DistanceSimilarityModule(SimilarityModuleBase):
```

rectools/models/nn/transformers/torch_backbone.py
```diff
@@ -168,11 +179,34 @@ def forward(
...
         """
         item_embs = self.item_model.get_all_embeddings()  # [n_items + n_item_extra_tokens, n_factors]
         session_embs = self.encode_sessions(sessions, item_embs)  # [batch_size, session_max_len, n_factors]
-        return item_embs, session_embs
+        logits = self.similarity_module(session_embs, item_embs, item_ids)
+        if last_n_items:
+            logits = logits[:, -last_n_items:, :]
+        return item_embs, session_embs, logits
+
+    def _recommend_u2i(
```
blondered, line 195: Do we really need it? We can just call `self.torch_model.similarity_module._recommend_u2i` instead

tests/models/nn/transformers/test_base.py
```diff
@@ -112,6 +113,7 @@ def test_save_load_for_unfitted_model(
         config = {
             "deterministic": True,
             "item_net_block_types": (IdEmbeddingsItemNet, CatFeaturesItemNet),
+            "similarity_module_type": SimilarityModuleDistance,
```
blondered, line 116: This should not be in the tests. Right now you need it only because the default value is not set. Let's set default instead

rectools/models/nn/transformers/base.py
```diff
@@ -237,13 +250,15 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-locals
         item_net_constructor_type: tp.Type[ItemNetConstructorBase] = SumOfEmbeddingsConstructor,
         pos_encoding_type: tp.Type[PositionalEncodingBase] = LearnableInversePositionalEncoding,
         lightning_module_type: tp.Type[TransformerLightningModuleBase] = TransformerLightningModule,
+        similarity_module_type: tp.Type[SimilarityModuleBase] = SimilarityModuleBase,
```
blondered, line 253:
```suggestion
        similarity_module_type: tp.Type[SimilarityModuleBase] = DistanceSimilarityModule,
```

rectools/models/nn/transformers/bert4rec.py
```diff
@@ -256,6 +257,8 @@ class BERT4RecModel(TransformerModelBase[BERT4RecModelConfig]):
         Type of data preparator used for dataset processing and dataloader creation.
     lightning_module_type : type(TransformerLightningModuleBase), default `TransformerLightningModule`
         Type of lightning module defining training procedure.
+    similarity_module_type : type(SimilarityModuleBase), default `SimilarityModuleBase`
```
blondered, line 260:
```suggestion
    similarity_module_type : type(SimilarityModuleBase), default `DistanceSimilarityModule`
```

rectools/models/nn/transformers/bert4rec.py
```diff
@@ -320,6 +326,7 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-locals
         transformer_layers_type: tp.Type[TransformerLayersBase] = PreLNTransformerLayers,
         data_preparator_type: tp.Type[TransformerDataPreparatorBase] = BERT4RecDataPreparator,
         lightning_module_type: tp.Type[TransformerLightningModuleBase] = TransformerLightningModule,
+        similarity_module_type: tp.Type[SimilarityModuleBase] = SimilarityModuleBase,
```
blondered, line 329:
```suggestion
        similarity_module_type: tp.Type[SimilarityModuleBase] = DistanceSimilarityModule,
```

rectools/models/nn/transformers/sasrec.py
```diff
@@ -336,6 +337,8 @@ class SASRecModel(TransformerModelBase[SASRecModelConfig]):
         Type of data preparator used for dataset processing and dataloader creation.
     lightning_module_type : type(TransformerLightningModuleBase), default `TransformerLightningModule`
         Type of lightning module defining training procedure.
+    similarity_module_type : type(SimilarityModuleBase), default `SimilarityModuleBase`
```
blondered, line 340:
```suggestion
    similarity_module_type : type(SimilarityModuleBase), default `DistanceSimilarityModule`
```

rectools/models/nn/transformers/sasrec.py
```diff
@@ -399,6 +405,7 @@ def __init__(  # pylint: disable=too-many-arguments, too-many-locals
         transformer_layers_type: tp.Type[TransformerLayersBase] = SASRecTransformerLayers,  # SASRec authors net
         data_preparator_type: tp.Type[TransformerDataPreparatorBase] = SASRecDataPreparator,
         lightning_module_type: tp.Type[TransformerLightningModuleBase] = TransformerLightningModule,
+        similarity_module_type: tp.Type[SimilarityModuleBase] = SimilarityModuleBase,
```
blondered, line 408:
```suggestion
        similarity_module_type: tp.Type[SimilarityModuleBase] = DistanceSimilarityModule,
```

tests/models/nn/transformers/test_sasrec.py
```diff
@@ -981,12 +1024,13 @@ def get_reco(model: SASRecModel) -> pd.DataFrame:
 
         model_1 = model.from_config(initial_config)
         reco_1 = get_reco(model_1)
-        config_1 = model_1.get_config(simple_types=simple_types)
+        config_1 = model_1.get_config(mode="pydantic", simple_types=False)
```
blondered, line 1027: Will it raise en error with `simple_types=simple_types`? Should pass (here and below)

tests/models/nn/transformers/test_bert4rec.py
```diff
@@ -837,6 +852,7 @@ def initial_config(self) -> tp.Dict[str, tp.Any]:
             "item_net_constructor_kwargs": None,
             "pos_encoding_kwargs": None,
             "lightning_module_kwargs": None,
+            "similarity_module_kwargs": {"dist": "dot"},
```
blondered, line 855: We should have default kwargs anywhere. But we do have default distance in DistanceSimilarityModule, so there is no need for this one here

tests/models/nn/transformers/test_bert4rec.py
```diff
@@ -876,6 +892,8 @@ def test_get_config(
                 "data_preparator_type": "rectools.models.nn.transformers.bert4rec.BERT4RecDataPreparator",
                 "lightning_module_type": "rectools.models.nn.transformers.lightning.TransformerLightningModule",
                 "get_val_mask_func": "tests.models.nn.transformers.utils.leave_one_out_mask",
+                "similarity_module_type": "rectools.models.nn.transformers.similarity.SimilarityModuleDistance",
+                "similarity_module_kwargs": {"dist": "dot"},
```
blondered, line 896: And here

rectools/models/nn/transformers/similarity.py
```diff
@@ -0,0 +1,110 @@
...
+    """Similarity module distance."""
+
+    dist_available: tp.List[str] = ["dot", "cosine"]
+    # dist_available_values: tp.List[int] = [Distance.DOT.value, Distance.COSINE.value]
+    epsilon_cosine_dist: float = 1e-8
+
+    def __init__(self, dist: str = "dot") -> None:
```
blondered, line 50:
```suggestion
    def __init__(self, distance: str = "dot") -> None:
```

rectools/models/nn/transformers/similarity.py
```diff
@@ -0,0 +1,110 @@
...
+        super().__init__()
+        if dist not in self.dist_available:
+            raise ValueError("`dist` can only be either `dot` or `cosine`.")
+
+        self.dist = dist
```
blondered, line 55:
```suggestion
        self.distance = Distance(distance)
```
Should work? Then we don't need an `if` in `_recommend_u2i`

rectools/models/nn/transformers/torch_backbone.py
```diff
@@ -168,11 +179,34 @@ def forward(
...
         Returns
         -------
-        (torch.Tensor, torch.Tensor)
+        (torch.Tensor, torch.Tensor, torch.Tensor)
         """
         item_embs = self.item_model.get_all_embeddings()  # [n_items + n_item_extra_tokens, n_factors]
         session_embs = self.encode_sessions(sessions, item_embs)  # [batch_size, session_max_len, n_factors]
-        return item_embs, session_embs
+        logits = self.similarity_module(session_embs, item_embs, item_ids)
+        if last_n_items:
```
blondered, line 191: I know I was the one to propose it, but could we please make a `forward` without an if? Let `forward` return all embeddings. And let's make a separate method that calls `forward` and returns only `-last_n_items` logits.
We are starting to think about ONNX and we now try to avoid `if` in `TorchBackbone`

rectools/models/nn/transformers/torch_backbone.py
```diff
@@ -168,11 +179,34 @@ def forward(
...
+            logits = logits[:, -last_n_items:, :]
+        return item_embs, session_embs, logits
```
blondered, line 193: Let's return logits only

rectools/models/nn/transformers/lightning.py
```diff
@@ -149,15 +151,17 @@ def training_step(self, batch: tp.Dict[str, torch.Tensor], batch_idx: int) -> to
         """Training step."""
         x, y, w = batch["x"], batch["y"], batch["yw"]
         if self.loss == "softmax":
-            logits = self._get_full_catalog_logits(x)
+            _, _, logits = self.torch_model(sessions=x)
```
blondered, line 154: This is just a suggestion. But to clean up all of our modifications code we can do the following:

- create a property `requires_negatives`. True for softmax, False for negative losses, None for unknown losses.
- Get logits depending on the property with common code for all losses.
- Then call just `if` for each lossname and calculated.

Or even cleaner: reuse TorchRanker logic `_get_scorer` logic from here: https://github.com/MobileTeleSystems/RecTools/blob/c5dc5f793e3727aade308fcdb3563f0e4c3d5fb7/rectools/models/rank/rank_torch.py#L178
We can create `_get_loss_calculator` which returns a flag for requiring negatives and a function to call to calculate loss.

What do you think?

rectools/models/nn/transformers/torch_backbone.py
blondered: Also if we do the refactoring for `if` logic in calculating losses like I proposed below, we can skip implementing `-last_n_items` here. Because it will be quite easy to do in lightning module after refactoring.

rectools/models/nn/transformers/lightning.py
In48semenov: - About reusing TorchRanker:
We should put the following methods `_calc_softmax_loss`, `_calc_bce_loss`, `_calc_gbce_loss` + `_get_reduced_overconfidence_logits` and `_calc_custom_loss_outputs` in `TransformerLightningModuleBase` so that you don't create another `__init__` in `TransformerLightningModule`. But we can't tell the difference between `bce` and `gbce` without an additional if, since the negatives are fed to the input of the gbce loss.

So I think we won't get away from if in these methods, so I consider the first one to be a cleaner option

rectools/models/nn/transformers/similarity.py
In48semenov: Doesn't work
`TypeError: __call__() missing 1 required positional argument: 'value'`

It can be work as `Distance(1)`.

tests/models/nn/transformers/test_sasrec.py
In48semenov: I forgot removed this after "debug"
Returned to previous version because it works.

rectools/models/nn/transformers/torch_backbone.py
In48semenov: Fixed

tests/models/nn/transformers/test_base.py
In48semenov: Fixed

tests/models/nn/transformers/test_bert4rec.py
In48semenov: Fixed

tests/models/nn/transformers/test_bert4rec.py
In48semenov: Fixed

rectools/models/nn/transformers/torch_backbone.py
In48semenov: Fixed

rectools/models/nn/transformers/torch_backbone.py
In48semenov: Fixed

rectools/models/nn/transformers/lightning.py
In48semenov: I selected option with properties `requires_negatives`

rectools/models/nn/transformers/similarity.py
blondered: Let's update there:
```
class Distance(str, Enum):
    """Distance metric"""

    DOT = "dot"  # Bigger value means closer vectors
    COSINE = "cosine"  # Bigger value means closer vectors
    EUCLIDEAN = "euclidean"  # Smaller value means closer vectors
```

rectools/models/nn/transformers/lightning.py
```diff
@@ -186,21 +206,29 @@ def validation_step(self, batch: tp.Dict[str, torch.Tensor], batch_idx: int) ->
...
+        if self.requires_negatives is not None:
+            logits = logits[:, -1:, :]
+
         outputs = {}
         if self.loss == "softmax":
-            logits = self._get_full_catalog_logits(x)[:, -1:, :]
             outputs["loss"] = self._calc_softmax_loss(logits, y, w)
             outputs["logits"] = logits.squeeze()
         elif self.loss == "BCE":
             negatives = batch["negatives"]
```
blondered, line 226: Should drop this line

rectools/models/nn/transformers/similarity.py
```diff
@@ -0,0 +1,109 @@
+import typing as tp
```
blondered, line 1: Please update changelog

rectools/models/nn/transformers/similarity.py
```diff
@@ -0,0 +1,109 @@
...
+        ui_csr_for_filter: tp.Optional[sparse.csr_matrix],
+    ) -> InternalRecoTriplet:
+        """Recommend to users."""
+        raise NotImplementedError()
+
+
+class DistanceSimilarityModule(SimilarityModuleBase):
+    """Distandce similarity module."""
+
+    dist_available: tp.List[str] = ["dot", "cosine"]
```
blondered, line 45:
```suggestion
    dist_available: tp.List[Distance] = [Distance.DOT, Distance.COSINE]
```
(After refactoring Distance)

rectools/models/nn/transformers/similarity.py
```diff
@@ -0,0 +1,109 @@
...
+    def _get_embeddings_norm(self, embeddings: torch.Tensor) -> torch.Tensor:
+        embedding_norm = torch.norm(embeddings, p=2, dim=1).unsqueeze(dim=1)
+        embeddings = embeddings / torch.max(embedding_norm, self.epsilon_cosine_dist)
+        return embeddings
+
+    def forward(
+        self, session_embs: torch.Tensor, item_embs: torch.Tensor, item_ids: tp.Optional[torch.Tensor] = None
+    ) -> torch.Tensor:
+        """Forward pass to get logits."""
+        if self.distance == "cosine":
```
blondered, line 76:
```suggestion
        if self.distance == Distance.COSINE:
```
After refactoring Distance

rectools/models/nn/transformers/similarity.py
```diff
@@ -0,0 +1,109 @@
...
+            session_embs = self._get_embeddings_norm(session_embs)
+            item_embs = self._get_embeddings_norm(item_embs)
+
+        if item_ids is None:
+            logits = self._get_full_catalog_logits(session_embs, item_embs)
+        else:
+            logits = self._get_pos_neg_logits(session_embs, item_embs, item_ids)
+        return logits
```
blondered, line 84:
```suggestion
        if item_ids is None:
            return self._get_full_catalog_logits(session_embs, item_embs)
        return self._get_pos_neg_logits(session_embs, item_embs, item_ids)
```

tests/models/nn/transformers/test_base.py
blondered: It is still in many places which makes me think tests would break without it :)
Could you please remove in cases where we just had
```
config = {
  "deterministic": True,
  "item_net_block_types": (IdEmbeddingsItemNet, CatFeaturesItemNet)
}
```

Actually, even "item_net_block_types" are not needed there. We forgot to remove them

rectools/models/nn/transformers/lightning.py
blondered: Seems good to me

tests/models/nn/transformers/test_base.py
In48semenov: Sorry, I forgot about other `"similarity_module_type"` in `test_base.py`
Also removed `"item_net_block_types"`

rectools/models/nn/transformers/similarity.py
```diff
@@ -40,19 +39,18 @@ def _recommend_u2i(
         raise NotImplementedError()
 
 
-class SimilarityModuleDistance(SimilarityModuleBase):
-    """Similarity module distance."""
+class DistanceSimilarityModule(SimilarityModuleBase):
+    """Distandce similarity module."""
```
blondered, line 43:
```suggestion
    """Distance similarity module."""
```

rectools/models/nn/transformers/data_preparator.py
```diff
@@ -168,6 +168,19 @@ def _process_features_for_id_map(
         full_feature_values = np.vstack([extra_token_feature_values, sorted_features.values])
         return DenseFeatures.from_iterables(values=full_feature_values, names=raw_features.names)
 
+    def _filter_train_interactions(self, interactions: pd.DataFrame) -> pd.DataFrame:
+        """Filter train interactions."""
+        train_interactions = interactions.copy()
```
blondered, line 173: Can we skip making a copy?

rectools/models/nn/transformers/lightning.py
```diff
@@ -148,16 +164,20 @@ def on_train_start(self) -> None:
     def training_step(self, batch: tp.Dict[str, torch.Tensor], batch_idx: int) -> torch.Tensor:
         """Training step."""
         x, y, w = batch["x"], batch["y"], batch["yw"]
+
+        if self.requires_negatives:
```
blondered, line 168: I think we can move this common logic in train and val steps to a separate method `get_logits(batch)` after `get_loss_calculator` is introduced.

rectools/models/nn/transformers/data_preparator.py
In48semenov: Removed

rectools/models/nn/transformers/lightning.py
In48semenov: I made `get_loss_calculator`, `get_bacth_logits(batch)` and changed the default value of n_negatives to 0.

rectools/models/nn/transformers/lightning.py
```diff
@@ -83,6 +85,7 @@ def __init__(
         self.data_preparator = data_preparator
         self.lr = lr
         self.loss = loss
+        self.loss_calculator, self.requires_negatives = self.get_loss_calculator()
```
blondered, line 88: Can we please split `loss_calculator` logic and `requires_negatives` logic?
We can now finally make it all clean with negatives requirement. We have a very tricky solution in data preparator initialization: https://github.com/MobileTeleSystems/RecTools/blob/f1583c5127de16760fef237733181c919f2a3a64/rectools/models/nn/transformers/base.py#L320C16-L320C78
We just hardcode "softmax" loss check there and it is not good.

In our new decision we can make it in the correct way all together: if we implement `requires_negatives(loss: str) -> bool` as a static method of `LightningModuleBase` then we can call it in data preparator initialization as `requires_negatives = self.lightning_module_type.requires_negatives(self.loss)`

Could we please do this as a last thing here? :)

In48semenov: That's a great point.
Fixed.

rectools/models/rank/rank_torch.py
```diff
@@ -194,8 +196,8 @@ def _euclid_score(self, user_embs: torch.Tensor, item_embs: torch.Tensor) -> tor
         return torch.cdist(user_embs.unsqueeze(0), item_embs.unsqueeze(0)).squeeze(0)
 
     def _cosine_score(self, user_embs: torch.Tensor, item_embs: torch.Tensor) -> torch.Tensor:
-        user_embs = user_embs / torch.norm(user_embs, p=2, dim=1).unsqueeze(dim=1)
-        item_embs = item_embs / torch.norm(item_embs, p=2, dim=1).unsqueeze(dim=1)
+        user_embs = user_embs / torch.max(torch.norm(user_embs, p=2, dim=1).unsqueeze(dim=1), self.epsilon_cosine_dist)
```
blondered, line 199: Move epsilon to correct device

rectools/models/nn/transformers/similarity.py
```diff
@@ -0,0 +1,107 @@
...
+        self, session_embs: torch.Tensor, item_embs: torch.Tensor, item_ids: torch.Tensor
+    ) -> torch.Tensor:
+        pos_neg_embs = item_embs[item_ids]  # [batch_size, session_max_len, len(item_ids), n_factors]
+        # [batch_size, session_max_len,len(item_ids)]
+        logits = (pos_neg_embs @ session_embs.unsqueeze(-1)).squeeze(-1)
+        return logits
+
+    def _get_embeddings_norm(self, embeddings: torch.Tensor) -> torch.Tensor:
+        embedding_norm = torch.norm(embeddings, p=2, dim=1).unsqueeze(dim=1)
+        embeddings = embeddings / torch.max(embedding_norm, self.epsilon_cosine_dist)
```
blondered, line 69: Move epsilon to correct device

rectools/models/rank/rank_torch.py
In48semenov: fixed.

rectools/models/nn/transformers/similarity.py
In48semenov: Fixed

blondered (approved):

PR #276, opened 2025-04-01
Embedding out dimension

spirinamayya:
Added embedding output dimension to Item nets

### DIFF
--- a/rectools/models/nn/item_net.py
+++ b/rectools/models/nn/item_net.py
@@ -449,6 +449,11 @@ class ItemNetConstructorBase(ItemNetBase):
         """
         raise NotImplementedError()
 
+    @property
+    def out_dim(self) -> int:
+        """Return item net constructor output dimension."""
+        raise NotImplementedError()
+
 
 class SumOfEmbeddingsConstructor(ItemNetConstructorBase):
     """
@@ -482,3 +487,8 @@ class SumOfEmbeddingsConstructor(ItemNetConstructorBase):
             item_emb = self.item_net_blocks[idx_block](items)
             item_embs.append(item_emb)
         return torch.sum(torch.stack(item_embs, dim=0), dim=0)
+
+    @property
+    def out_dim(self) -> int:
+        """Return item net constructor output dimension."""
+        return self.item_net_blocks[0].out_dim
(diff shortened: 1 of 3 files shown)


tests/models/nn/test_item_net.py
```diff
@@ -295,6 +301,15 @@ def test_warns_when_dataset_schema_categorical_features_are_none(self) -> None:
             """
         )
 
+    @pytest.mark.parametrize("n_factors", ((2), (10)))
+    def test_out_dim(self, dataset_item_features: Dataset, n_factors: int) -> None:
+        cat_item_embeddings = CatFeaturesItemNet.from_dataset(
+            dataset_item_features, n_factors=n_factors, dropout_rate=0.5
+        )
+        if cat_item_embeddings is not None:
```
blondered, line 309: You can't do `if` in test. It can skip the actual check that is needed while something may be broken inside. 
In tests it is necessary to do ignores for linters in such cases.

rectools/models/nn/item_net.py
```diff
@@ -317,6 +327,11 @@ def from_dataset_schema(
         n_items = dataset_schema.items.n_hot
         return cls(n_factors, n_items, dropout_rate)
 
+    @property
```
blondered, line 330: Let's also implement `out_dim` for `SumOfEmbeddingsConstructor` as out_dim of the first block. We might need it in the future.

rectools/models/nn/item_net.py
```diff
@@ -46,6 +46,11 @@ def get_all_embeddings(self) -> torch.Tensor:
         """Return item embeddings."""
         raise NotImplementedError()
 
+    @property
+    def out_dim(self) -> int:
```
blondered, line 50: Please update changelog

tests/models/nn/test_item_net.py
spirinamayya: fixed

rectools/models/nn/item_net.py
spirinamayya: fixed

rectools/models/nn/item_net.py
```diff
@@ -449,6 +449,11 @@ def forward(self, items: torch.Tensor) -> torch.Tensor:
         """
         raise NotImplementedError()
 
+    @property
+    def out_dim(self) -> int:
+        """Return item net constructor output dimension."""
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
spirinamayya: fixed

blondered (approved):

PR #306, opened 2026-04-24
Adding UniSRec model implemented on lightweight class hierarchy with pytorch preprocessing 

TOPAPEC:
## Motivation

- Introduce a standalone `rectools.fast_transformers` package for sequential recommendation models (UniSRec, FlatSASRec) that work directly with PyTorch tensors, without relying on the `Dataset`/pandas pipeline. The main motivation is speeding up and simplifying extendability of the NN recommenders. Also aiming to simplify production use of rectools models by reducing boilerplates and making dataflow inside the model more straightforward. 
- Provide fast sequence preprocessing and embedding-alignment utilities needed for efficient training and benchmarking workflows.
- Faster training on ml-20m fast_transformers/ vs rectools native pipeline - 900sec vs 1400sec - 10 epochs 

## Description

- Added a new `rectools.fast_transformers` package with:
  - `metrics` — GPU-computed ranking metrics - much faster at scale compared to pandas ones. 
  - `net` — FlatSASRec implementation;
  - `preprocessing` — vectorized `build_sequences`, `align_embeddings`, and `SequenceBatchDataset` - also much faster at scale. 
  - `unisrec` — model network, Lightning module, high-level model API, ONNX export helpers, and demo docs.
- Added unit tests under `tests/fast_transformers` covering:
  - metrics;
  - preprocessing and sequence building;
  - FlatSASRec and UniSRec networks;
  - Lightning wrapper behavior;
  - model fit/predict flows and ONNX export roundtrips.
- Updated project/config artifacts to integrate the new package:
  - flake8 per-file ignores for the new modules;
  - repository metadata/docs artifacts (`.gitignore`, `CHANGELOG.md`);
  - benchmark artifacts under `benchmark/...`.

## Typing/Lint note

- Added a targeted `# type: ignore[import-untyped]` for `requests` in `benchmark/compare_sasrec_unisrec.py` so `mypy` passes in the current lint environment without changing dependency policy.

## Testing

- New fast_transformers tests are included under `tests/fast_transformers` and are part of the standard test suite structure.


rectools/fast_transformers/gpu_data.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def build_sequences(
+    user_ids: torch.Tensor,
+    item_ids: torch.Tensor,
+    timestamps: torch.Tensor,
+    max_len: int,
+    min_interactions: int = 2,
+    device: str = "cuda",
+    id_mapping: str = "dense",
```
feldlime: Better to use Literal for such things

rectools/fast_transformers/preprocessing/sequence_data.py
```diff
@@ -0,0 +1,151 @@
...
+
+def build_sequences(
+    user_ids: torch.Tensor,
+    item_ids: torch.Tensor,
+    timestamps: torch.Tensor,
+    max_len: int,
+    min_interactions: int = 2,
+    device: str = "cuda",
+    id_mapping: str = "dense",
+) -> tp.Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
```
feldlime, line 23: Please add extensive docstrings for all the public method, especially for those supposed to be used stand-alone. Here it's especially important since you're returning 4 tensors and user doesn't understand their meaning. Also good to add examples

.gitignore
```diff
@@ -95,4 +95,9 @@ benchmark_results/
 *.dat
 
 # CatBoost
-catboost_info/
\ No newline at end of file
+catboost_info/
+
+# Dev artifacts
+training_folder/
```
feldlime: a bit weird name, can we remove it?

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+- `UniSRec` network and `UniSRecModel` — sequential recommender with pretrained text embeddings (e.g. Qwen) and a learnable PCA/BN adaptor. Three-phase training: (1) SASRec warm-up on ID embeddings, (2) adaptor-only with frozen transformer, (3) full fine-tune on pretrained embeddings. Configurable losses (softmax, BCE, gBCE, sampled_softmax), optimizers (Adam, AdamW), cosine warmup scheduler, early stopping, checkpoint save/load. `UniSRecModel.fit()` accepts raw `(user_ids, item_ids, timestamps)` tensors
+- `rank_topk()` utility for batched top-k scoring with CSR-based viewed-item filtering and item whitelist support
+- `align_embeddings()` for mapping pretrained embedding matrices to internal item ID order
+- `GPUBatchDataset` and `make_dataloader()` — lightweight torch Dataset/DataLoader wrappers for sequence training data
+- Configurable FFN blocks in `UniSRec`: `conv1d` (original paper), `linear_gelu`, `linear_relu` with adjustable expansion factor
+- Tests for all `fast_transformers` submodules (143 tests)
```
feldlime: We normally don't add anything that doesn't affect user directly to the changelog, so not much sense to write about the tests

benchmark/compare_sasrec_unisrec.py
feldlime, line 1: please put this script and the report to a subfolder in the `benchmark` folder

rectools/fast_transformers/gpu_data.py
```diff
@@ -0,0 +1,151 @@
...
+    elif id_mapping == "hash":
+        positions = hash_item_ids(idx, n_items)
+        aligned[positions[valid]] = pretrained[idx[valid]]
+    else:
+        raise ValueError(f"Unknown id_mapping: {id_mapping}. Use 'dense' or 'hash'")
+
+    return aligned
+
+
+class GPUBatchDataset(TorchDataset):
```
feldlime: I'm not sure the name reflect the purpose

- why GPU?
- what does Batch mean?

It also sounds quite "universal" even though I'd say it's more task-specific

rectools/fast_transformers/gpu_data.py
```diff
@@ -0,0 +1,151 @@
...
+            batch = self.transform(batch)
+        return batch
+
+
+def make_dataloader(
+    x: torch.Tensor,
+    y: torch.Tensor,
+    batch_size: int,
+    shuffle: bool = True,
+    transform: tp.Optional[tp.Callable] = None,
```
feldlime: I'd recommend to add **kwargs here to cover different parameters of data loader 

On the other side I'm not sure it makes much sense to wrap 2 function calls in a separate function

rectools/fast_transformers/ranking.py
```diff
@@ -0,0 +1,80 @@
+"""Batch top-k ranking with optional viewed-item filtering."""
+
+import typing as tp
+
+import numpy as np
+import torch
+from scipy import sparse
+
+
+def rank_topk(
```
feldlime: Sorry, I'm too lazy to check, could you please describe why do we need it given that we have `TorchRanker`? Could we reuse the code?

rectools/fast_transformers/preprocessing/sequence_data.py
TOPAPEC: done

rectools/fast_transformers/preprocessing/sequence_data.py
```diff
@@ -0,0 +1,151 @@
...
+    user_ids = user_ids.to(device)
+    item_ids = item_ids.to(device)
+    timestamps = timestamps.to(device)
```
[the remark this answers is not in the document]
TOPAPEC: resolved inability to set device and fixed unneeded transfer

rectools/fast_transformers/gpu_data.py
TOPAPEC: Done, updated the type to Literal

.gitignore
TOPAPEC: Done, removed that entry

CHANGELOG.md
TOPAPEC: Done, removed test-only changelog notes

benchmark/compare_sasrec_unisrec.py
TOPAPEC: Done, moved benchmark artifacts into a dedicated subfolder.

rectools/fast_transformers/gpu_data.py
TOPAPEC: Done, renamed to a more task-specific and explicit name. preprocessing-> sequence_data

rectools/fast_transformers/gpu_data.py
TOPAPEC: I simplified the wrapper and added kwargs passthrough where needed

rectools/fast_transformers/ranking.py
TOPAPEC: this path handles sequence-specific filtering efficiently; TorchRanker reuse is limited here. I removed this file, this is 5-lines logic, handled on user side https://github.com/TOPAPEC/RecTools/blob/65ef231ed0c9cd1f444cddd06fc265e80509a11a/benchmark/compare_sasrec_unisrec.py#L174

rectools/fast_transformers/gpu_data.py
```diff
@@ -0,0 +1,151 @@
...
+    item_ids = item_ids.to(device)
+    timestamps = timestamps.to(device)
+
+    unique_items = torch.unique(item_ids)
+    n_unique = len(unique_items)
+
+    if id_mapping == "dense":
+        _, item_inv = torch.unique(item_ids, return_inverse=True)
+        internal_items = item_inv + 1
+    elif id_mapping == "hash":
```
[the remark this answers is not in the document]
TOPAPEC: Done, replaced with a single torch.unique(..., return_inverse=True) call

rectools/fast_transformers/unisrec/model.py
```diff
@@ -0,0 +1,458 @@
...
+            dropout=self.dropout,
+            adaptor_dropout=self.adaptor_dropout,
+            adaptor_type=self.adaptor_type,
+            use_adaptor_ffn=self.use_adaptor_ffn,
+            ffn_type=self.ffn_type,
+            ffn_expansion=self.ffn_expansion,
+        )
+
+        train_dl = make_dataloader(x, y, batch_size=self.batch_size, shuffle=True)
+
```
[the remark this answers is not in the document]
TOPAPEC: Done, added an explicit empty-input check with a clear ValueError

rectools/fast_transformers/unisrec_model.py
```diff
@@ -0,0 +1,458 @@
...
+        if self.id_mapping == "hash":
+            n_items = len(self._unique_items)
+            known = torch.isin(external_ids, self._unique_items)
+            result = torch.zeros_like(external_ids)
+            result[known] = hash_item_ids(external_ids[known], n_items)
+            return result
+
+        lookup = {int(v): i + 1 for i, v in enumerate(self._unique_items.tolist())}
+        return torch.tensor([lookup.get(int(x), 0) for x in external_ids.tolist()], dtype=torch.long)
+
```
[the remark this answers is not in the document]
TOPAPEC: Done, switched to vectorized tensor mapping and preserved the original device

rectools/fast_transformers/unisrec_model.py
```diff
@@ -0,0 +1,458 @@
...
+            ffn_type=self.ffn_type,
+            ffn_expansion=self.ffn_expansion,
+        )
+
+        train_dl = make_dataloader(x, y, batch_size=self.batch_size, shuffle=True)
+
+        val_dl = None
+        if self.patience is not None:
+            val_y_last = y[:, -1:]
+            val_dl = make_dataloader(x, val_y_last, batch_size=self.batch_size, shuffle=False)
```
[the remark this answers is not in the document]
TOPAPEC: Done, wired it into both train and validation DataLoaders

tests/fast_transformers/test_unisrec_net.py
```diff
@@ -0,0 +1,115 @@
...
+
+class TestPaddingInvariance:
+    def test_same_input_same_output(self, net: UniSRec) -> None:
+        net.eval()
+        x_a = torch.tensor([[0, 0, 0, 5, 10]])
+        x_b = torch.tensor([[0, 0, 0, 5, 10]])
+        with torch.no_grad():
+            e_a = net.encode_last(x_a, use_id=False)
+            e_b = net.encode_last(x_b, use_id=False)
+        torch.testing.assert_close(e_a, e_b)
```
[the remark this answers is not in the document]
TOPAPEC: Done, clarified test naming/intent to avoid implying a different invariance check

rectools/fast_transformers/ranking.py
```diff
@@ -0,0 +1,80 @@
...
+        end = min(start + batch_size, n_users)
+        scores = user_embs[start:end] @ item_embs.T  # (batch, M)
+
+        if filter_csr is not None:
+            batch_csr = filter_csr[start:end]
+            if whitelist is not None:
+                batch_csr = batch_csr[:, whitelist]
+            viewed_mask = torch.tensor(batch_csr.toarray(), dtype=torch.bool, device=device)
+            scores[viewed_mask] = -float("inf")
+
```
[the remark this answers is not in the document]
TOPAPEC: Done, verified current inference path is predict_topk (torch path); no active rank_topk densify path in the current tree.

rectools/fast_transformers/unisrec/model.py
```diff
@@ -0,0 +1,458 @@
...
+        self
+        """
+        x, y, unique_items, unique_users = build_sequences(
+            user_ids,
+            item_ids,
+            timestamps,
+            max_len=self.session_max_len,
+            min_interactions=self.train_min_user_interactions,
+            id_mapping=self.id_mapping,
+        )
```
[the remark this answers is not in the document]
TOPAPEC: fixed, now device can be chosen

tests/fast_transformers/test_net.py
```diff
@@ -0,0 +1,46 @@
...
+    def test_padding_invariance(self, net: FlatSASRec) -> None:
+        """Different left-padding should produce same last-position embedding."""
+        net.eval()
+        # Same content should produce identical output
+        x_a = torch.tensor([[0, 0, 0, 5, 10]])
+        x_b = torch.tensor([[0, 0, 0, 5, 10]])
+        with torch.no_grad():
+            e_a = net.encode_last(x_a)
+            e_b = net.encode_last(x_b)
+        torch.testing.assert_close(e_a, e_b)
```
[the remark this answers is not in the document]
TOPAPEC: Done, clarified test naming/intent to avoid implying a different invariance check.

rectools/fast_transformers/unisrec/net.py
```diff
@@ -0,0 +1,303 @@
...
+# keep old name as alias
+FeedForward = FeedForwardConv1d
+
+
+def make_ffn(n_factors: int, ffn_type: str, expansion: int, dropout: float) -> nn.Module:
+    """Create a feed-forward block.
+
+    Parameters
+    ----------
+    ffn_type : ``"conv1d"`` | ``"linear_gelu"`` | ``"linear_relu"``
```
feldlime, line 49: Could you please add all the parameters and the return

rectools/fast_transformers/metrics.py
```diff
@@ -0,0 +1,149 @@
...
+        if k > k_max:
+            raise ValueError(f"k={k} exceeds topk_ids width {k_max}")
+        top = topk_ids[:, :k]
+        results[f"HR@{k}"] = hitrate_at_k(top, targets).item()
+        results[f"NDCG@{k}"] = ndcg_at_k(top, targets, log_base=log_base).item()
+        results[f"MRR@{k}"] = mrr_at_k(top, targets).item()
+    return results
+
+
+def _log(base: int) -> float:
```
feldlime: it's used only once and it's a one-liner - are you sure you need it?

CHANGELOG.md
```diff
@@ -6,6 +6,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## [Unreleased]
+
+### Added
+- `rectools.fast_transformers` module — standalone transformer-based sequential recommenders that work directly with torch tensors, bypassing the `Dataset`/pandas pipeline. GPU-native sequence building via `build_sequences()` gives ~30x preprocessing speedup over `SASRecDataPreparator` on ML-20M
```
feldlime: Please add links to the PR (see examples below)

setup.cfg
```diff
@@ -43,10 +43,15 @@ exclude = .venv
 docstring-convention = numpy
 ignore = D205,D400,D105,D100,E203,W503
 per-file-ignores =
-    tests/*: D100,D101,D102,D103,D104
+    tests/*: D100,D101,D102,D103,D104,N806
```
feldlime: why do we need to ignore N806? Can't we don't violate it?

Is it violated in many places? (If not, better to have local ignores)

If absolutely necessary, please specify it only for the fast_transformers subfolder

setup.cfg
```diff
@@ -43,10 +43,15 @@ exclude = .venv
...
+    benchmark/*: D100,D101,D102,D103,D104
     rectools/models/nn/dssm.py: D101,D102,N812
     rectools/dataset/torch_datasets.py: D101,D102
     rectools/models/implicit_als.py: N806
+    rectools/fast_transformers/net.py: D102,N806
+    rectools/fast_transformers/unisrec/lightning.py: D102,D401,N812
+    rectools/fast_transformers/unisrec/model.py: D102
+    rectools/fast_transformers/unisrec/net.py: D102,N806
```
feldlime: 1. It's not the best way to specify rules per file - unless really needed
2. Please don't ignore D102 and D401 - all public methods must have a docstring, and in the correct style
3. I'd also really prefer not to violate N812
4. N806 is not so critical

feldlime (changes requested):

rectools/fast_transformers/unisrec/model.py
```diff
@@ -0,0 +1,510 @@
...
+            batch_size=self.batch_size,
+            shuffle=True,
+            num_workers=self.dataloader_num_workers,
+        )
+
+        val_dl = None
+        if self.patience is not None:
+            val_y_last = y[:, -1:]
+            val_dl = DataLoader(
+                SequenceBatchDataset(x, val_y_last, transform=neg_transform),
```
feldlime, line 349: Early stopping is measured on the **training data**: `val_dl` wraps the same `x` as `train_dl` (just with `y` reduced to the last position). So `val_loss` decreases monotonically as the model overfits, and `patience` either never triggers or triggers for unrelated reasons (noise / regularization stochasticity). Either accept an explicit validation split in `fit()` (e.g. `val_user_ids/val_item_ids/val_timestamps`, or hold out the last interaction per user), or document that `patience` is effectively a train-loss plateau heuristic, not real early stopping.

feldlime (commented):
A few more findings that haven't been raised yet — mostly around `UniSRecModel.fit/save/load/predict/export` lifecycle and the embedding-alignment helper. Nothing here blocks the architecture, but several would bite users in practice.

rectools/fast_transformers/unisrec/model.py
```diff
@@ -0,0 +1,510 @@
...
+                "unique_users": self._unique_users,
+                "n_items": len(self._unique_items),
+            },
+            path,
+        )
+
+    def load_checkpoint(self, path: tp.Union[str, Path], device: tp.Optional[str] = None) -> None:
+        if device is None:
+            device = "cuda" if torch.cuda.is_available() else "cpu"
+        ckpt = torch.load(path, map_location=device, weights_only=False)  # nosec B614
```
feldlime: `torch.load(path, weights_only=False)` is an arbitrary-pickle deserialization — silencing bandit with `# nosec B614` hides the risk, it doesn't address it. The checkpoint here only stores tensors and a Python int, so this should work with `weights_only=True` once `unique_items`/`unique_users` are saved as plain tensors (already the case). Please switch to `weights_only=True` (PyTorch is making this the default in 2.6+ anyway), and we avoid shipping a model that will RCE on a malicious checkpoint.

rectools/fast_transformers/unisrec/model.py
```diff
@@ -0,0 +1,510 @@
...
+
+    def save_checkpoint(self, path: tp.Union[str, Path]) -> None:
+        assert self._net is not None and self._unique_items is not None
+        torch.save(
+            {
+                "net": self._net.state_dict(),
+                "unique_items": self._unique_items,
+                "unique_users": self._unique_users,
+                "n_items": len(self._unique_items),
+            },
```
feldlime, line 381: `save_checkpoint` stores `state_dict` but no architecture hyperparameters (`n_factors`, `n_blocks`, `n_heads`, `session_max_len`, `adaptor_type`, `use_adaptor_ffn`, `ffn_type`, `ffn_expansion`, ...). `load_checkpoint` rebuilds `UniSRec` from `self.*` — i.e. whatever the user passed to the new `UniSRecModel(...)` constructor before calling `load_checkpoint`. If they don't pass the same arch, `load_state_dict` either fails loudly (shape mismatch) or, worse, silently loads partial weights. Save the relevant hparams in the checkpoint dict and restore them in `load_checkpoint`, or at minimum add a `@classmethod from_checkpoint(cls, path)` that constructs the model from saved hparams.

rectools/fast_transformers/unisrec/model.py
```diff
@@ -0,0 +1,510 @@
...
+        assert self._net is not None, "Model not fitted or loaded"
+        net = self._net
+        was_training = net.training
+        net.eval()
+        device = next(net.parameters()).device
+        h = net.encode_last(input_ids.to(device))
+        item_embs = net.project_all()
+        scores = h @ item_embs.T
+        scores[:, 0] = float("-inf")
+        top_scores, top_ids = scores.topk(k, dim=1)
```
feldlime: `predict_topk` returns **internal** item IDs (1..n_items), but the user gives external IDs everywhere else (in `fit`, in `map_item_ids`). The model already holds the mapping (`self._unique_items`), so making every caller post-process `top_ids` is friction and an easy source of subtle bugs (e.g. computing HR/NDCG against external targets while comparing to internal IDs). Either return external IDs by default and offer an `internal=True` flag, or change the return type to a small dataclass that documents the ID space.

rectools/fast_transformers/unisrec/model.py
```diff
@@ -0,0 +1,510 @@
...
+        dummy = torch.zeros(1, 5, dtype=torch.long, device=device)
+
+        torch.onnx.export(
+            net,
+            (dummy,),
+            str(encoder_path),
+            input_names=["input_ids"],
+            output_names=["hidden"],
+            opset_version=opset_version,
+        )
```
feldlime: The ONNX export is traced with `dummy = torch.zeros(1, 5, ...)` and no `dynamic_axes`, so the resulting graph only accepts inputs of shape `(1, 5)`. In practice anyone exporting for serving wants dynamic batch and (often) dynamic sequence length up to `session_max_len`. Suggest passing `dynamic_axes={"input_ids": {0: "batch", 1: "seq"}, "hidden": {0: "batch", 1: "seq"}}` (and similar for the `project_all` wrapper — its output is `(n_items+1, D)`, so dynamic `n_items` if you ever expect to re-export after refit). Also worth using `session_max_len` for the dummy length so positional embeddings are exercised at their full range during tracing.

rectools/fast_transformers/preprocessing/sequence_data.py
```diff
@@ -0,0 +1,173 @@
...
+    device = pretrained.device
+    idx = unique_items.long().to(device)
+    valid = (idx >= 0) & (idx < pretrained.shape[0])
+
+    if pretrained.ndim == 2:
+        aligned = torch.zeros(n_items + 1, pretrained.shape[1], device=device)
+    else:
+        aligned = torch.zeros(n_items + 1, pretrained.shape[1], pretrained.shape[2], device=device)
+
+    aligned[1:][valid] = pretrained[idx[valid]]
```
feldlime, line 167: Two issues here:

1. `valid = (idx >= 0) & (idx < pretrained.shape[0])` silently zeros out unknown item embeddings — making them indistinguishable from the padding row (also zeros). If `pretrained` was built from a stale catalog, the model will silently treat missing items as padding rather than warning. Suggest raising (or at least logging) when `(~valid).any()`.

2. `torch.zeros(n_items + 1, ...)` defaults to `float32`, so if `pretrained` is `bfloat16`/`float16`/`float64` the alignment silently changes dtype (and then everything downstream is fp32). Use `dtype=pretrained.dtype, device=pretrained.device`.

rectools/fast_transformers/unisrec/model.py
```diff
@@ -0,0 +1,510 @@
...
+        was_training = net.training
+        net.eval()
+        device = next(net.parameters()).device
+        h = net.encode_last(input_ids.to(device))
+        item_embs = net.project_all()
+        scores = h @ item_embs.T
+        scores[:, 0] = float("-inf")
+        top_scores, top_ids = scores.topk(k, dim=1)
+        if was_training:
+            net.train()
```
feldlime: `net.eval() ... if was_training: net.train()` isn't exception-safe — if `encode_last` raises (CUDA OOM, shape mismatch, ...) the network is left in `eval()` mode for the rest of the session, which silently disables dropout on subsequent `fit()`. Same pattern in `export_to_onnx`. Wrap in `try`/`finally` or use a small context manager.

rectools/fast_transformers/__init__.py
```diff
@@ -0,0 +1,23 @@
+"""Fast Transformers: flat sequential recommenders without ItemNet hierarchy."""
+
+from .metrics import compute_metrics, hitrate_at_k, mrr_at_k, ndcg_at_k
+from .net import FlatSASRec, SASRecBlock
+from .preprocessing import SequenceBatchDataset, align_embeddings, build_sequences
+from .unisrec import UniSRec, UniSRecLightning, UniSRecModel
```
feldlime: Top-level package exposes both `UniSRec` (the `nn.Module`) and `UniSRecModel` (the high-level trainer). Two classes whose only difference is a `Model` suffix is genuinely confusing on import (`from rectools.fast_transformers import UniSRec` vs `UniSRecModel`). Same goes for `FlatSASRec` (network) sitting next to `UniSRecModel` (trainer) in the same namespace. Suggest renaming the networks to `UniSRecNet`/`FlatSASRecNet` (matching `lightning.py` / `model.py` vocabulary), or only re-exporting the high-level `*Model` classes here and keeping nets reachable via `rectools.fast_transformers.unisrec.net`.

rectools/fast_transformers/unisrec/lightning.py
```diff
@@ -0,0 +1,207 @@
...
+        if self.loss_name == "BCE":
+            return self._bce_loss(logits, mask)
+        if self.loss_name == "gBCE":
+            return self._gbce_loss(logits, mask)
+
+        raise ValueError(f"Unknown loss: {self.loss_name}")
+
+    def _full_softmax_loss(self, hidden: torch.Tensor, labels: torch.Tensor) -> torch.Tensor:
+        all_emb = self._get_all_embs()
+        logits = hidden @ all_emb.T
```
feldlime, line 112: `_full_softmax_loss` scores against `project_all()`, which uses **variant 0** of `frozen_emb` (deterministic). The other losses score against `_get_item_embs(...) = _adapt_score(_sample_frozen(...))`, which randomly samples variants during training. For multi-variant pretrained embeddings this means the training signal is fundamentally different for `softmax` vs `BCE`/`gBCE`/`sampled_softmax` — softmax never sees variant augmentation. Either route the full-softmax path through a sampled-variant projection too, or document that variant augmentation only applies to negative-sampling losses.

rectools/fast_transformers/net.py
```diff
@@ -0,0 +1,175 @@
...
+        # timeline_mask: zero out padding positions to prevent NaN from attention
+        timeline_mask = (x != self.PADDING_IDX).unsqueeze(-1).float()  # (B, L, 1)
+        attn_mask = self._causal_mask(L, x.device)
+        key_padding_mask = x == self.PADDING_IDX
+
+        for block in self.blocks:
+            h = h * timeline_mask
+            h = block(h, attn_mask=attn_mask, key_padding_mask=key_padding_mask)
+        h = h * timeline_mask
+        h = self.final_ln(h)
```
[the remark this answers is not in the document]
TOPAPEC: Didn't catch it. I think we should propagate nans and make sure user knows that nans are occurring in their code and not silently handle them.

rectools/fast_transformers/metrics.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Returns
+    -------
+    Tensor (scalar)
+        Mean NDCG across users.
+    """
+    k = topk_ids.shape[1]
+    hits = (topk_ids == targets.unsqueeze(1)).float()  # (B, K)
+    ranks = torch.arange(1, k + 1, device=topk_ids.device, dtype=torch.float)
+    discounts = 1.0 / torch.log(ranks + 1) * (1.0 / _log(log_base))
```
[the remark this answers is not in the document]
TOPAPEC: Skip, just cosmetics

rectools/fast_transformers/unisrec/model.py
```diff
@@ -0,0 +1,510 @@
...
+        self.n_items = n_items
+        self.n_negatives = n_negatives
+
+    def __call__(self, batch: tp.Dict[str, torch.Tensor]) -> tp.Dict[str, torch.Tensor]:
+        y = batch["y"]
+        negs = torch.randint(1, self.n_items + 1, (*y.shape, self.n_negatives), device=y.device)
+        # Resample positions where negative == positive
+        collisions = negs == y.unsqueeze(-1)
+        if collisions.any():
+            negs[collisions] = torch.randint(1, self.n_items + 1, (int(collisions.sum()),), device=y.device)
```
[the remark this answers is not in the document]
TOPAPEC: Skip, too low probability of that happening - won't affect training

rectools/fast_transformers/unisrec/lightning.py
TOPAPEC: done

feldlime (dismissed):

feldlime (approved):

